PR #113, opened 2020-04-14
feat(edge-compute): deploy stack through docker

### LINKED ISSUE #112: Add the ability to deploy a stack via Docker binary
Add support for Docker binary management to deploy a local stack.

chiptus:
fix #112


agent.go
```diff
@@ -134,6 +135,14 @@ type (
 	Scheduler interface {
 		Schedule(schedules []Schedule) error
 	}
+
+	// EdgeStackManager is a service to manager edge stacks
```
deviantony, line 139:
```suggestion
	// EdgeStackManager is a service to manage Edge stacks
```

cmd/agent/main.go
```diff
@@ -107,6 +109,17 @@ func main() {
 		if err != nil {
 			log.Fatalf("[ERROR] [main,edge,rtunnel] [message: Unable to start agent in Edge mode] [error: %s]", err)
 		}
+
+		edgeStackManager, err = exec.NewEdgeStackManager(options.DockerBinaryPath)
+		if err != nil {
+			log.Fatalf("[ERROR] [main,edge,stack] [message: Unable to start stack manager] [error: %s]", err)
+		}
+
+		err = edgeStackManager.Login()
```
deviantony, line 118: No need to login during startup.

exec/stack.go
```diff
@@ -0,0 +1,99 @@
...
+	manager := &EdgeStackManager{
+		binaryPath: binaryPath,
+	}
+
+	return manager, nil
+}
+
+// Login executes the docker login command against a list of registries (including DockerHub).
+func (manager *EdgeStackManager) Login() error {
+	// dockerhub *portainer.DockerHub, registries []portainer.Registry
```
deviantony, line 29: The comment block can be entirely removed for now, it might be different from the portainer implementation.

Replace with a // Not implemented yet. comment

exec/stack_test.go
```diff
@@ -0,0 +1,59 @@
+package exec
```
deviantony, line 1: I don't think we should bothered with unit testing on that one.

os/options.go
```diff
@@ -44,6 +45,7 @@ func (parser *EnvOptionParser) Options() (*agent.Options, error) {
 		EdgeInactivityTimeout: agent.DefaultEdgeSleepInterval,
 		EdgeInsecurePoll:      false,
 		LogLevel:              agent.DefaultLogLevel,
+		DockerBinaryPath:      os.Getenv(EnvKeyDockerBinaryPath),
```
deviantony, line 48: No need to provide the Docker binary path as an option. It will be hardcoded to `/app/docker` or (`/app/docker.exe` on windows)

chiptus: sounds good. Where should I hardcode it? into the stack manager?

deviantony: Should be hardcoded in the domain (`agent.go` with the other constants).

PR #120, opened 2020-04-26
feat(tunnel): start polling only if leader

### LINKED ISSUE #117: [edge compute] Update the agent to trigger the poll request on the leader only
When running inside a Swarm cluster, the agent should only trigger the short poll request on the node that is the leader of the Swarm cluster.

This will allow us to avoid spamming poll requests from all the nodes in the cluster and to make sure that we're always reaching out to a manager node.

We'll leverage Swarm cluster failover as well.

chiptus:
fix  #117

### DIFF
--- a/cmd/agent/main.go
+++ b/cmd/agent/main.go
@@ -108,7 +108,7 @@ func main() {
 			log.Fatalf("[ERROR] [main,edge,rtunnel] [message: Unable to create tunnel operator] [error: %s]", err)
 		}
 
-		err := enableEdgeMode(tunnelOperator, clusterService, infoService, options)
+		err := enableEdgeMode(tunnelOperator, clusterService, options)
 		if err != nil {
 			log.Fatalf("[ERROR] [main,edge,rtunnel] [message: Unable to start agent in Edge mode] [error: %s]", err)
 		}
@@ -165,7 +165,7 @@ func startAPIServer(config *http.APIServerConfig) error {
 	return server.StartSecured()
 }
 
-func enableEdgeMode(tunnelOperator agent.TunnelOperator, clusterService agent.ClusterService, infoService agent.InfoService, options *agent.Options) error {
+func enableEdgeMode(tunnelOperator agent.TunnelOperator, clusterService agent.ClusterService, options *agent.Options) error {
 
 	edgeKey, err := retrieveEdgeKey(options, clusterService)
 	if err != nil {
--- a/cmd/agent/main.go
+++ b/cmd/agent/main.go
@@ -224,8 +224,6 @@ func enableEdgeMode(tunnelOperator agent.TunnelOperator, clusterService agent.Cl
 		for {
 			select {
 			case <-ticker.C:
-				log.Printf("[DEBUG] [main,edge,docker] [message: checking docker config and key]")
-
 				key := tunnelOperator.GetKey()
 				if key == "" {
 					continue
@@ -238,18 +236,19 @@ func enableEdgeMode(tunnelOperator agent.TunnelOperator, clusterService agent.Cl
 				}
 
 				if agentTags[agent.MemberTagEngineStatus] == agent.EngineStatusStandalone || agentTags[agent.MemberTagKeyIsLeader] == "1" {
-					log.Printf("[DEBUG] [main,edge,docker] [message: this is either a leader or a standalone agent, starting tunnel operator]")
 					err = tunnelOperator.Start()
 					if err != nil {
 						log.Printf("[ERROR] [main,edge,docker] [message: an error occured while starting poll] [error: %s]", err)
 					}
+
 				} else {
-					log.Printf("[DEBUG] [main,edge,docker] [message: this is neither a leader nor a standalone agent, starting tunnel operator]")
 					err = tunnelOperator.Stop()
 					if err != nil {
 						log.Printf("[ERROR] [main,edge,docker] [message: an error occured while stopping poll] [error: %s]", err)
 					}
 				}
+
+				log.Printf("[DEBUG] [main,edge,docker] [] [message: Docker runtime configuration check] [engine_status: %s] [leader_node: %s]", agentTags[agent.MemberTagEngineStatus], agentTags[agent.MemberTagKeyIsLeader] == "1")
 			}
 		}
 	}()
--- a/cmd/agent/main.go
+++ b/cmd/agent/main.go
@@ -187,15 +187,17 @@ func enableEdgeMode(tunnelOperator agent.TunnelOperator, clusterService agent.Cl
 		if err != nil {
 			return err
 		}
+
 		isLeaderNode, err := infoService.IsLeaderNode()
 		if err != nil {
 			return err
 		}
+
 		if isLeaderNode {
 			return tunnelOperator.Start()
 		}
 
-		ticker := time.NewTicker(time.Duration(5) * time.Second)
+		ticker := time.NewTicker(time.Duration(agent.DefaultSwarmLeaderCheckInterval) * time.Second)
 		go func() {
 			for {
 				select {
@@ -209,10 +211,6 @@ func enableEdgeMode(tunnelOperator agent.TunnelOperator, clusterService agent.Cl
 						tunnelOperator.Start()
 						ticker.Stop()
 					}
-					// case <-operator.refreshSignal:
-					// 	log.Println("[DEBUG] [http,edge,poll] [message: shutting down Portainer short-polling client]")
-					// 	ticker.Stop()
-					// 	return
 				}
 			}
 		}()
--- a/cmd/agent/main.go
+++ b/cmd/agent/main.go
@@ -91,11 +91,14 @@ func main() {
 		apiServerAddr := fmt.Sprintf("%s:%s", advertiseAddr, options.AgentServerPort)
 
 		operatorConfig := &tunnel.OperatorConfig{
-			APIServerAddr:     apiServerAddr,
-			EdgeID:            options.EdgeID,
-			PollFrequency:     agent.DefaultEdgePollInterval,
-			InactivityTimeout: options.EdgeInactivityTimeout,
-			InsecurePoll:      options.EdgeInsecurePoll,
+			APIServerAddr:        apiServerAddr,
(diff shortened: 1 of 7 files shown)


docker/docker.go
```diff
@@ -109,3 +109,20 @@ func (service *InfoService) GetServiceNameFromDockerEngine(containerName string)
...
+// IsLeaderNode returns true if the node is the leader node of the swarm cluster
+func (service *InfoService) IsLeaderNode() (bool, error) {
+	cli, err := client.NewClientWithOpts(client.FromEnv, client.WithVersion(agent.SupportedDockerAPIVersion))
+	if err != nil {
+		return false, err
+	}
+
+	defer cli.Close()
+
+	node, _, err := cli.NodeInspectWithRaw(context.Background(), "self")
```
chiptus, line 122: @deviantony would love your help here:
this fails on worker nodes, which should be ok because it fails also when using the cli. Maybe in this case we should just return false?
This also fails on manager nodes, because it doesn't recognize node `self`, but in the cli `docker node inspect self` works.

Maybe I should do it differently? list all nodes and find the leader?

chiptus:
I should test on standalone agent

cmd/agent/main.go
```diff
@@ -185,31 +187,70 @@ func enableEdgeMode(tunnelOperator agent.TunnelOperator, clusterService agent.Cl
...
+				if key == "" {
+					continue
+				}
+
+				agentTags, err := infoService.GetInformationFromDockerEngine()
+				if err != nil {
+					log.Printf("[ERROR] [main,edge,poll] [message: an error occured during docker config check] [error: %s]", err)
+					continue
+				}
+
```
deviantony, line 239: Add the following DEBUG statement here:

```
log.Printf("[DEBUG] [main,edge,docker] [] [message: Docker runtime configuration check] [engine_status: %s] [leader_node: %s]", agentTags[agent.MemberTagEngineStatus], agentTags[agent.MemberTagKeyIsLeader] == "1")
```

See comment above and below about removing other DEBUG statements.

cmd/agent/main.go
```diff
@@ -185,31 +187,70 @@ func enableEdgeMode(tunnelOperator agent.TunnelOperator, clusterService agent.Cl
...
+				if agentTags[agent.MemberTagEngineStatus] == agent.EngineStatusStandalone || agentTags[agent.MemberTagKeyIsLeader] == "1" {
+					log.Printf("[DEBUG] [main,edge,poll] [message: this is either a leader or a standalone agent, starting tunnel operator]")
+					err = tunnelOperator.Start()
+					if err != nil {
+						log.Printf("[ERROR] [main,edge,poll] [message: an error occured while starting poll] [error: %s]", err)
```
deviantony, line 244:
```suggestion
						log.Printf("[ERROR] [main,edge,poll] [message: an error occured while starting the short-poll process] [error: %s]", err)
```

cmd/agent/main.go
```diff
@@ -185,31 +187,70 @@ func enableEdgeMode(tunnelOperator agent.TunnelOperator, clusterService agent.Cl
...
+					}
+				} else {
+					log.Printf("[DEBUG] [main,edge,poll] [message: this is neither a leader nor a standalone agent, starting tunnel operator]")
+					err = tunnelOperator.Stop()
+					if err != nil {
+						log.Printf("[ERROR] [main,edge,poll] [message: an error occured while stopping poll] [error: %s]", err)
```
deviantony, line 250:
```suggestion
						log.Printf("[ERROR] [main,edge,poll] [message: an error occured while stopping the short-poll process] [error: %s]", err)
```

cmd/agent/main.go
```diff
@@ -185,31 +187,70 @@ func enableEdgeMode(tunnelOperator agent.TunnelOperator, clusterService agent.Cl
 				return err
 			}
 		}
+	} else {
+		log.Println("[DEBUG] [main,edge] [message: Edge key not specified. Serving Edge UI]")
+		edgeServer := http.NewEdgeServer(tunnelOperator, clusterService)
 
-		return tunnelOperator.Start()
```
deviantony, line 181: Since the tunnel operator is not started here anymore the following debug statement should be updated.

```
		log.Println("[DEBUG] [main,edge] [message: Edge key available. Starting tunnel operator.]")
```

to:

```
		log.Println("[DEBUG] [main,edge] [message: Edge key found in environment. Associating Edge key to cluster.]")
```

cmd/agent/main.go
```diff
@@ -185,31 +187,70 @@ func enableEdgeMode(tunnelOperator agent.TunnelOperator, clusterService agent.Cl
...
+				log.Printf("[INFO] [main,edge,http] [message: Shutting down Edge UI server as no key was specified after %d minutes]", agent.DefaultEdgeSecurityShutdown)
+				edgeServer.Shutdown()
+			}
+		}()
+	}
 
-	go func() {
-		timer1 := time.NewTimer(agent.DefaultEdgeSecurityShutdown * time.Minute)
-		<-timer1.C
+	loopIntervalFrequency, err := time.ParseDuration(agent.DefaultConfigCheckInterval)
```
deviantony, line 216:
```suggestion
	runtimeCheckFrequency, err := time.ParseDuration(agent.DefaultConfigCheckInterval)
```

cmd/agent/main.go
```diff
@@ -185,31 +187,70 @@ func enableEdgeMode(tunnelOperator agent.TunnelOperator, clusterService agent.Cl
...
+	if err != nil {
+		return err
+	}
 
-		if !tunnelOperator.IsKeySet() {
-			log.Printf("[INFO] [main,edge,http] [message: Shutting down Edge UI server as no key was specified after %d minutes]", agent.DefaultEdgeSecurityShutdown)
-			edgeServer.Shutdown()
+	ticker := time.NewTicker(time.Duration(loopIntervalFrequency))
```
deviantony, line 221:
```suggestion
	ticker := time.NewTicker(loopIntervalFrequency)
```

cmd/agent/main.go
```diff
@@ -162,7 +164,7 @@ func startAPIServer(config *http.APIServerConfig) error {
 	return server.StartSecured()
 }
 
-func enableEdgeMode(tunnelOperator agent.TunnelOperator, clusterService agent.ClusterService, options *agent.Options) error {
+func enableEdgeMode(tunnelOperator agent.TunnelOperator, clusterService agent.ClusterService, infoService agent.InfoService, options *agent.Options) error {
```
deviantony, line 167: The `enableEdgeMode` function starts to look complex/hard to read, I would advise to extract it in multiple functions.

Here is a proposition of the final result for that function (related functions not provided):

```go
func enableEdgeMode(tunnelOperator agent.TunnelOperator, clusterService agent.ClusterService, infoService agent.InfoService, options *agent.Options) error {
	edgeKey, err := retrieveEdgeKey(options, clusterService)
	if err != nil {
		return err
	}

	if edgeKey != "" {
		log.Println("[DEBUG] [main,edge] [message: Edge key available. Starting tunnel operator.]")

		err := associateEdgeKey(tunnelOperator, clusterService, edgeKey)
		if err != nil {
			return err
		}
	} else {
		log.Println("[DEBUG] [main,edge] [message: Edge key not specified. Serving Edge UI]")
		serveEdgeUI(tunnelOperator, clusterService, options)
	}

	err = startRuntimeConfigCheckProcess(tunnelOperator, infoService)
	if err != nil {
		return err
	}

	return nil
}
```

cmd/agent/main.go
```diff
@@ -185,31 +187,70 @@ func enableEdgeMode(tunnelOperator agent.TunnelOperator, clusterService agent.Cl
...
+				log.Printf("[DEBUG] [main,edge,poll] [message: checking docker config and key]")
+
+				key := tunnelOperator.GetKey()
+				if key == "" {
+					continue
+				}
+
+				agentTags, err := infoService.GetInformationFromDockerEngine()
+				if err != nil {
+					log.Printf("[ERROR] [main,edge,poll] [message: an error occured during docker config check] [error: %s]", err)
```
deviantony, line 236:
```suggestion
					log.Printf("[ERROR] [main,edge,poll] [message: an error occured during Docker runtime configuration check] [error: %s]", err)
```

docker/docker.go
```diff
@@ -109,3 +117,13 @@ func (service *InfoService) GetServiceNameFromDockerEngine(containerName string)
 
 	return containerInspect.Config.Labels[serviceNameLabel], nil
 }
+
+// IsLeaderNode returns true if the node is the leader node of the swarm cluster
```
deviantony, line 121: No need to add a comment on an unexported function.

docker/docker.go
```diff
@@ -41,12 +41,20 @@ func (service *InfoService) GetInformationFromDockerEngine() (map[string]string,
 	info[agent.MemberTagKeyNodeName] = dockerInfo.Name
 
 	if dockerInfo.Swarm.NodeID == "" {
```
deviantony, line 43: To make this block clearer I would advise extracting its content into two unexported functions:

```go
	if dockerInfo.Swarm.NodeID == "" {
		getStandaloneInformation(info)
	} else {
		err := getSwarmInformation(info)
		if err != nil {
			return nil, err
		}
	}
```

The content of the `isLeaderNode` function could be integrated into the `getSwarmInformation` function.

cmd/agent/main.go
```diff
@@ -292,11 +263,90 @@ func parseOptions() (*agent.Options, error) {
...
+					}
+
+				} else {
+					err = tunnelOperator.Stop()
+					if err != nil {
+						log.Printf("[ERROR] [main,edge,docker] [message: an error occured while stopping the short-poll process] [error: %s]", err)
+					}
+				}
+
+				log.Printf("[DEBUG] [main,edge,docker] [] [message: Docker runtime configuration check] [engine_status: %s] [leader_node: %b]", agentTags[agent.MemberTagEngineStatus], agentTags[agent.MemberTagKeyIsLeader] == "1")
```
deviantony, line 346: This log entry should probably be located right after the error check for `infoService.GetInformationFromDockerEngine()`.

PR #121, opened 2020-05-04
feat(edge-compute): add stack support

### LINKED ISSUE #115: [edge compute] - Update the agent to support Edge stacks
The Edge agent must be able to retrieve Edge stack related information as part of the periodic short poll request.

The agent must be able to store information related to Edge stacks (identifier, version number) in memory.

It must also be able to retrieve a stack file associated to an Edge stack and apply it.

If a stack is referenced in the memory of the agent but not exposed via short the short poll request anymore, the stack must be deleted.

chiptus:
closes #115


chiptus:
I'm not sure about the location of the EdgeStackManager and all of its types (maybe it needs to be under docker or edge instead of HTTP?)

http/portainerclient/portainer_client.go
```diff
@@ -0,0 +1,118 @@
+package portainerclient
```
deviantony, line 1: This file should be relocated inside the `client` package.

chiptus:
few things to think about:
1. dependency of tunnelOperator on edgeStackManager. I would prefer to remove it and then the edgeManager can include the functionality of the stackManager. Maybe we should do one of these:
   - separate polling status from polling for stacks status, then we can do polling for stacks in the edgeManager
  - move polling to the edge manager.
  - pass a callback to tunnelOperator that runs when stacks are found
2. in order to set key from the edgeUIServer we pass the manager itself to the server, which feels a bit circular too. Maybe a callback as well?

internal/edge/edge.go
```diff
@@ -0,0 +1,188 @@
...
+		<-timer1.C
+
+		if !manager.IsKeySet() {
+			log.Printf("[INFO] [main,edge,http] [message: Shutting down Edge UI server as no key was specified after %d minutes]", agent.DefaultEdgeSecurityShutdown)
+			edgeServer.Shutdown()
+		}
+	}()
+}
+
+func (manager *EdgeManager) startRuntimeConfigCheckProcess() error {
```
deviantony, line 121:
```suggestion
func (manager *EdgeManager) startEdgeBackgroundProcess() error {
```

internal/edge/key.go
```diff
@@ -0,0 +1,164 @@
...
+	return edgeKey, nil
+}
+
+func encodeKey(edgeKey *edgeKey) string {
+	keyInfo := fmt.Sprintf("%s|%s|%s|%s", edgeKey.PortainerInstanceURL, edgeKey.TunnelServerAddr, edgeKey.TunnelServerFingerprint, edgeKey.EndpointID)
+	encodedKey := base64.RawStdEncoding.EncodeToString([]byte(keyInfo))
+	return encodedKey
+}
+
+func (manager *EdgeManager) retrieveEdgeKey(edgeKey string) (string, error) {
```
deviantony, line 98: I think that this is not the responsability of the EdgeManager to retrieve the Edge key, this should be kept in the `main.go`.

http/tunnel/tunnel.go
```diff
@@ -122,13 +84,20 @@ func (operator *Operator) ResetActivityTimer() {
 // if needed as well as manage schedules.
 // The second loop will check for the last activity of the reverse tunnel and close the tunnel if it exceeds the tunnel
 // inactivity duration.
-func (operator *Operator) Start() error {
-	if operator.key == nil {
-		return errors.New("missing Edge key")
+func (operator *Operator) Start(portainerURL, endpointID, tunnelServerAddr, tunnelServerFingerprint string) error {
```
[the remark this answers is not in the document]
chiptus: so the tunnel operator (poll service) should be created when the key is set

internal/edge/stack.go
```diff
@@ -0,0 +1,242 @@
...
+		errorMessage = err.Error()
+	} else {
+		log.Printf("[DEBUG] [internal,edge,stack] [stack_identifier: %d] [message: stack deployed]", stack.ID, stack.Version)
+	}
+
+	manager.stacks[stack.ID] = stack
+
+	err = manager.httpClient.SetEdgeStackStatus(int(stack.ID), responseStatus, errorMessage)
+	if err != nil {
+		log.Printf("[ERROR] [internal,edge,stack] [message: failed setting edge stack status] [error: %v]", err)
```
deviantony, line 223:
```suggestion
		log.Printf("[ERROR] [internal,edge,stack] [message: unable to update Edge stack status] [error: %s]", err)
```

internal/edge/stack.go
```diff
@@ -0,0 +1,242 @@
...
+	}
+}
+
+func (manager *StacksManager) deleteStack(stack *edgeStack, stackName, stackFileLocation string) {
+	log.Printf("[DEBUG] [internal,edge,stack] [message: removing stack %d]", stack.ID)
```
deviantony, line 228:
```suggestion
	log.Printf("[DEBUG] [internal,edge,stack] [stack_identifier: %d] [message: removing stack]", stack.ID)
```

internal/edge/stack.go
```diff
@@ -0,0 +1,242 @@
...
+	err := filesystem.RemoveFile(stackFileLocation)
+	if err != nil {
+		log.Printf("[ERROR] [internal,edge,stack] [message: failed deleting edge stack file] [error: %v]", err)
```
deviantony, line 231:
```suggestion
		log.Printf("[ERROR] [internal,edge,stack] [message: unable to delete Edge stack file] [error: %s]", err)
```

internal/edge/stack.go
```diff
@@ -0,0 +1,242 @@
...
+		return
+	}
+
+	err = manager.dockerStackService.Remove(stackName)
+	if err != nil {
+		log.Printf("[ERROR] [internal,edge,stack] [message: failed removing stack] [error: %v]", err)
```
deviantony, line 237:
```suggestion
		log.Printf("[ERROR] [internal,edge,stack] [message: unable to remove stack] [error: %s]", err)
```

cmd/agent/main.go
```diff
@@ -87,41 +86,41 @@ func main() {
 		defer clusterService.Leave()
 	}
 
-	var tunnelOperator agent.TunnelOperator
-	var edgeStackManager agent.EdgeStackManager
+	edgeManager, err := edge.NewManager(options, advertiseAddr, clusterService, infoService)
```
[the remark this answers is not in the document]
chiptus: but then I need to pass them around to anywhere we're initializing the edge manager (HTTP handlers, edge UI server...)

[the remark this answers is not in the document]
deviantony: True, I missed that. Agree with you then.

agent.go
```diff
@@ -232,4 +220,6 @@ const (
 	EdgeKeyFile = "agent_edge_key"
 	// DockerBinaryPath is the path of the docker binary
 	DockerBinaryPath = "/app"
+	// EdgeStackFilesPath is the path where edge stack files are saved
+	EdgeStackFilesPath = "/tmp/edge_stacls"
```
deviantony, line 224:
```suggestion
	EdgeStackFilesPath = "/tmp/edge_stacks"
```

PR #126, opened 2020-05-10
feat(edge-stacks): use edge stack name for deployed name

### LINKED ISSUE #125: [edge-compute] - Rename deployed Edge stacks from edge_stackid to edge_stackname
Replace the `edge_stackid` deployment pattern with `edge_stackname`.

chiptus:
Closes #125

### DIFF
--- a/agent.go
+++ b/agent.go
@@ -27,6 +27,13 @@ type (
 		EdgeKeySet bool
 	}
 
+	// EdgeStackConfig represnts an Edge stack config
+	EdgeStackConfig struct {
+		Name        string
+		FileContent string
+		Prune       bool
+	}
+
 	// AgentMetadata is the representation of the metadata object used to decorate
 	// all the objects in the response of a Docker aggregated resource request.
 	Metadata struct {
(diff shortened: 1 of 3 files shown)


http/client/portainer_client.go
```diff
@@ -34,39 +34,40 @@ func NewPortainerClient(serverAddress, endpointID, edgeID string) *PortainerClie
...
 
 type stackConfigResponse struct {
+	Name string
 	StackFileContent string
 	Prune            bool
 }
 
 // GetEdgeStackConfig retrieves the configuration associated to an Edge stack
-func (client *PortainerClient) GetEdgeStackConfig(edgeStackID int) (string, bool, error) {
+func (client *PortainerClient) GetEdgeStackConfig(edgeStackID int) (string, string, bool, error) {
```
deviantony, line 43: 3 return values are OK in some cases but should be exceptional.

4+ should never happen, please introduce a structure that you'll use here instead.

agent.go
```diff
@@ -27,6 +27,13 @@ type (
 		EdgeKeySet bool
 	}
 
+	// EdgeStackConfig represnts an Edge stack config
+	EdgeStackConfig struct {
```
deviantony, line 31: Probably something that can be relocated into libedge in the future.

PR #137, opened 2020-05-27
fix(poll): init refresh signal on restart

### LINKED ISSUE #136: Changing poll interval cause the agent to poll multiple times
1. run portainer with an edge agent.
2. change the poll interval
3. see that poll logs appear twice (validated by creating a ticker id for each `startPollLoop`)

chiptus:
fix #136


deviantony:
Can you open a new 2.0 branch and rebase it against 2.0 ?

chiptus:
done

deviantony:
Ok so the 2.0 branch is not required anymore actually so I rebased it on develop.

PR #184, opened 2021-04-18
Clarify used base64 encoding in README

hhromic:
* The agent uses unpadded base64 encoding

The code base uses [`RawStdEncoding`](https://golang.org/pkg/encoding/base64/#pkg-variables) which uses unpadded base64 encoding.

I spent an unhealthy amount of time figuring out why my manually-encoded base64 Edge Key was not being recognised by the Portainer Agent until I realised it needed to be unpadded :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -118,7 +118,7 @@ to access the Edge UI again.
 
 ### Edge key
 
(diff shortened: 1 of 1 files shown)


deviantony (approved):
LGTM

deviantony:
Thanks for the clarification @hhromic :-)

PR #175, opened 2021-02-17
feat(edge): support edge stacks for standalone

chiptus:
fix [CE-208]

[CE-208]: https://portainer.atlassian.net/browse/CE-208

PR #186, opened 2021-05-13
 feat(edge): support edge stacks for kubernetes

chiptus:
needs to be merged with portainer PR (https://github.com/portainer/portainer-ee/pull/378)

blocked by #175

### DIFF
--- a/exec/kubernetes_deploy.go
+++ b/exec/kubernetes_deploy.go
@@ -31,7 +31,7 @@ func NewKubernetesDeployer(binaryPath string) *KubernetesDeployer {
 	}
 }
 
-// Deploy will deploy a Kubernetes manifest inside a specific namespace
+// Deploy will deploy a Kubernetes manifest inside the default namespace
 // it will use kubectl to deploy the manifest.
 // kubectl uses in-cluster config.
 func (deployer *KubernetesDeployer) Deploy(ctx context.Context, name string, filePaths []string, prune bool) error {
@@ -42,8 +42,6 @@ func (deployer *KubernetesDeployer) Deploy(ctx context.Context, name string, fil
 	stackFilePath := filePaths[0]
 
 	args := make([]string, 0)
-	// Specifying "--insecure-skip-tls-verify" make kubectl return error "default cluster has no server defined"
-	//args = append(args, "--insecure-skip-tls-verify")
 	args = append(args, "--namespace", "default")
 	args = append(args, "apply", "-f", stackFilePath)
 
--- a/exec/kubernetes_deploy.go
+++ b/exec/kubernetes_deploy.go
@@ -2,35 +2,91 @@ package exec
 
 import (
 	"bytes"
+	"context"
+	"errors"
 	"fmt"
-	"github.com/portainer/agent"
 	"os"
 	"os/exec"
 	"path"
 	"runtime"
 	"strings"
+
+	"github.com/portainer/agent"
 )
 
 // KubernetesDeployer represents a service to deploy resources inside a Kubernetes environment.
 type KubernetesDeployer struct {
-	binaryPath string
+	command string
 }
 
 // NewKubernetesDeployer initializes a new KubernetesDeployer service.
 func NewKubernetesDeployer(binaryPath string) *KubernetesDeployer {
+	command := path.Join(binaryPath, "kubectl")
+	if runtime.GOOS == "windows" {
+		command = path.Join(binaryPath, "kubectl.exe")
+	}
+
 	return &KubernetesDeployer{
-		binaryPath: binaryPath,
+		command: command,
 	}
 }
 
 // Deploy will deploy a Kubernetes manifest inside a specific namespace
 // it will use kubectl to deploy the manifest.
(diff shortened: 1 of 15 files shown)


exec/kubernetes_deploy.go
```diff
@@ -2,35 +2,90 @@ package exec
...
+		command = path.Join(binaryPath, "kubectl.exe")
+	}
+
 	return &KubernetesDeployer{
-		binaryPath: binaryPath,
+		command: command,
 	}
 }
 
 // Deploy will deploy a Kubernetes manifest inside a specific namespace
```
cmenginnz: Only inside the default namespace?

exec/kubernetes_deploy.go
```diff
@@ -2,35 +2,90 @@ package exec
...
 // it will use kubectl to deploy the manifest.
-func (deployer *KubernetesDeployer) Deploy(token, data string, namespace string) ([]byte, error) {
-	command := path.Join(deployer.binaryPath, "kubectl")
-	if runtime.GOOS == "windows" {
-		command = path.Join(deployer.binaryPath, "kubectl.exe")
+// kubectl uses in-cluster config.
+func (deployer *KubernetesDeployer) Deploy(name string, filePaths []string, prune bool) error {
+	if len(filePaths) == 0 {
```
cmenginnz, line 35: This function always runs kubectl with in-cluster config which has admin permission. It's a security hold.
As DeployRawConfig does,  it should run kubectl with the token from the request if there is any.

I'm not sure portainer allows non-admin users to deploy stacks or not. If the answer is not, the above comment is not necessary. But it's always good to use the token from request IMO.

exec/kubernetes_deploy.go
```diff
@@ -2,35 +2,90 @@ package exec
...
+	_, err := cmd.Output()
+	if err != nil {
+		return fmt.Errorf("failed deploying kubernetes stack %w: %s", err, stderr.String())
 	}
 
+	return nil
+}
+
+func (deployer *KubernetesDeployer) Remove(name string, filePaths []string) error {
+	if len(filePaths) == 0 {
```
cmenginnz, line 55: Same security concern as Deploy method.

exec/kubernetes_deploy.go
```diff
@@ -2,35 +2,90 @@ package exec
...
+	}
+
+	return nil
+}
+
+// DeployRawConfig will deploy a Kubernetes manifest inside a specific namespace
+// it will use kubectl to deploy the manifest and receives a raw config.
+// kubectl uses in-cluster config.
+func (deployer *KubernetesDeployer) DeployRawConfig(token, config string, namespace string) ([]byte, error) {
 	args := make([]string, 0)
```
cmenginnz: DeployRawConfig and Deploy share many common codes. It would be nice if pull them into a sharing method.

mcpacino (changes requested):

internal/edge/stack.go
```diff
@@ -54,16 +53,25 @@ const (
...
 type StackManager struct {
-	isSwarm            *bool
-	stacks             map[edgeStackID]*edgeStack
-	stopSignal         chan struct{}
-	dockerStackService agent.DockerStackService
-	portainerURL       string
-	endpointID         string
-	isEnabled          bool
-	httpClient         *client.PortainerClient
+	engineStatus *engineStatus
```
cmenginnz: It looks like that using type engineStatus but pointer type *engineStatus is enough and simpler here.
And naming engineType makes more sense to me.

mcpacino (changes requested):

exec/kubernetes_deploy.go
chiptus: yes

exec/kubernetes_deploy.go
chiptus: edge stacks are deployable by admin only. `Deploy` isn't called by an http request so there's no token.

internal/edge/stack.go
```diff
@@ -53,18 +53,18 @@ const (
...
+	_ engineType = iota
+	engineTypeDockerStandalone
+	engineTypeDockerSwarm
+	engineTypeKubernetes
 )
 
 // StackManager represents a service for managing Edge stacks
 type StackManager struct {
-	engineStatus *engineStatus
+	engineType   *engineType
```
cmenginnz: why need pointer here?

mcpacino (changes requested):

internal/edge/stack.go
chiptus: to be able to have a null value (engine is not set yet)

cmenginnz: It's an ok option if it needs to be a null value.

But personally I would like to explicit define the null value as a enum number. For example:
const (
	engineTypeNone engineType = iota
	engineTypeDockerStandalone
	engineTypeDockerSwarm
	engineTypeKubernetes
)

mcpacino (approved):

internal/edge/stack.go
chiptus: done

PR #199, opened 2021-07-22
fix(dockerhub): dockerhub ratelimit issues EE-1035

hookenz:
Fixes [EE-1035](https://portainer.atlassian.net/browse/EE-1035)

### DIFF
--- a/http/handler/dockerhub/dockerhub_status.go
+++ b/http/handler/dockerhub/dockerhub_status.go
@@ -115,14 +115,16 @@ func getDockerHubLimits(httpClient *http.Client, token string) (*dockerhubStatus
 		return nil, errors.New("failed fetching dockerhub limits")
 	}
 
+	// don't return an error if these headers are missing.  Dockerhub Pro accounts
+	// do not have these headers.  Returning an error adds confusing debug into the browser window.
 	rateLimit, err := parseNumericHeader(resp.Header, "RateLimit-Limit")
 	if err != nil {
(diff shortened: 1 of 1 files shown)


sbusso (approved):

http/handler/dockerhub/dockerhub_status.go
```diff
@@ -115,14 +115,16 @@ func getDockerHubLimits(httpClient *http.Client, token string) (*dockerhubStatus
...
 	rateLimit, err := parseNumericHeader(resp.Header, "RateLimit-Limit")
 	if err != nil {
-		return nil, fmt.Errorf("Failed fetching RateLimit-Limit header: %w", err)
+		return nil, nil
 	}
 
 	rateLimitRemaining, err := parseNumericHeader(resp.Header, "RateLimit-Remaining")
 	if err != nil {
-		return nil, fmt.Errorf("Failed fetching RateLimit-Remaining header: %w", err)
+		return nil, nil
```
DimaSalakhov, line 127: with these changes, dockerhub status call will return a json response `200 OK  null` (if I'm not mistaken)
Is that an expected behaviour?
Not a tricky question, genuinely curious.

hookenz: Just a way to return no data and supress the error as it's not really an error.
It was either that or return "{ }".  I went for the nil approach.

itsconquest (approved):
LGTM

PR #202, opened 2021-07-26
feat(edge) enable signature checking for edge agent  EE-743

mcpacino:


http/handler/key/handler.go
```diff
@@ -26,9 +26,9 @@ func NewHandler(notaryService *security.NotaryService, edgeManager *edge.Manager
 	}
 
 	h.Handle("/key",
-		notaryService.DigitalSignatureVerification(httperror.LoggerHandler(h.keyInspect))).Methods(http.MethodGet)
+		httperror.LoggerHandler(h.keyInspect)).Methods(http.MethodGet)
```
chiptus, line 29: why was this removed?

chiptus (approved):
Looks good. There's one question about the removal of notaryService.

http/handler/key/handler.go
cmenginnz: `/key` endpoint is used by other agents who do not have the signature, so remove the verification for it.

PR #236, opened 2022-02-08
feat(nomad) EE-1866 nomad as an environment

cmenginnz:
close EE-1866

### DIFF
--- a/cmd/agent/main.go
+++ b/cmd/agent/main.go
@@ -45,6 +45,7 @@ func main() {
 	var dockerInfoService agent.DockerInfoService
 	var advertiseAddr string
 	var kubeClient *kubernetes.KubeClient
+	var nomadConfig agent.NomadConfig
 
 	// !Generic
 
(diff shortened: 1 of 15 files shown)


internal/edge/edge.go
```diff
@@ -163,6 +163,43 @@ func (manager *Manager) startEdgeBackgroundProcessOnKubernetes(runtimeCheckFrequ
 	return nil
 }
 
+func (manager *Manager) startEdgeBackgroundProcessOnNomad(runtimeCheckFrequency time.Duration) error {
```
chiptus: looks like a copy of the k8s edge process. why not merge both, and receive (engineType) argument?

cmd/agent/main.go
```diff
@@ -158,8 +159,28 @@ func main() {
...
+			log.Fatalf("[ERROR] [main,nomad] [message: Unable to retrieve local IP associated to the agent] [error: %s]", err)
+		}
+
+		nomadConfig.NomadAddr = goos.Getenv(agent.NomadAddrEnvVarName)
+		if nomadConfig.NomadAddr == "" {
+			log.Fatalf("[ERROR] [main,nomad] [message: Unable to retrieve environment variable NOMAD_ADDR]")
+		}
 
+		nomadConfig.NomadToken = goos.Getenv(agent.NomadTokenEnvVarName)
+		if nomadConfig.NomadToken == "" {
```
chiptus, line 175: @mcpacino 

I think nomad token should be optional, no?

also, to get env vars, you might consider use kingpin (and also receive it as a flag). see https://github.com/portainer/agent/blob/ec007609438be768aa2cb9e5a6668a32874b6459/os/options.go#L38

internal/edge/edge.go
cmenginnz: I did it on purpose as do not mix Kube and Nomad things together.

cmd/agent/main.go
cmenginnz: yes, there is a ticket to make it optional

PR #252, opened 2022-02-20
feat(nomad): EE-2508 proxy requests to nomad server

cmenginnz:


cmd/agent/main.go
```diff
@@ -169,6 +170,15 @@ func main() {
 		if err != nil {
 			log.Fatalf("[ERROR] [main,nomad] [message: Unable to retrieve local IP associated to the agent] [error: %s]", err)
 		}
+
+		nomadConfig.NomadAddr = goos.Getenv(agent.NomadAddrEnvVarName)
```
deviantony, line 174: Where are these variables coming from? Set by the user when deploying the agent or available in the environment as part of a Nomad installation?

cmd/agent/main.go
```diff
@@ -169,6 +170,15 @@ func main() {
...
+		if nomadConfig.NomadAddr == "" {
+			log.Fatalf("[ERROR] [main,nomad] [message: Unable to retrieve environment variable NOMAD_ADDR]")
+		}
+
+		nomadConfig.NomadToken = goos.Getenv(agent.NomadTokenEnvVarName)
```
deviantony, line 179: This one is not mandatory?

cmd/agent/main.go
cmenginnz: These two environment variables are set as part of a Nomad installation.

cmd/agent/main.go
cmenginnz: yes, it should be mandatory. Fixed.

ArrisLee (approved):
LGTM

PR #253, opened 2022-02-21
fix(dev): deploy endpoint locally

chiptus:
deploying locally didn't work. 

tested the changes by deploying:
- locally
- standalone with ip
- swarm with two ips


deviantony (approved):
LGTM

PR #258, opened 2022-02-24
chore(server): refactor server Start function [EE-2606]

cheloRydel:
Closes [EE-2606](https://portainer.atlassian.net/browse/EE-2606)


http/server.go
```diff
@@ -152,26 +112,45 @@ func (server *APIServer) StartSecured(edgeMode bool) error {
...
+		}
+
+		server.edgeManager.ResetActivityTimer()
+
+		next.ServeHTTP(w, r)
+	})
+}
+
+func (server *APIServer) nonEdgeHandler(next http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
```
chiptus: isn't this whole function the same as `return next`? why do we need this?

http/server.go
```diff
@@ -88,55 +76,27 @@ func (server *APIServer) StartUnsecured(edgeMode bool) error {
 		EdgeManager:          server.edgeManager,
 		KubeClient:           server.kubeClient,
 		KubernetesDeployer:   server.kubernetesDeployer,
-		Secured:              false,
+		Secured:              !edgeMode,
 		ContainerPlatform:    server.containerPlatform,
 	}
 
-	var h http.Handler = handler.NewHandler(config)
+	var httpHandler http.Handler = handler.NewHandler(config)
```
chiptus: I know it was like this, but looks better IMO 
```suggestion
	httpHandler := handler.NewHandler(config)
```

chiptus (approved):
LGTM. see comments for suggestions

http/server.go
cheloRydel: yes you're right, I think it can be simplified by just deleting the function

deviantony (approved):
LGTM

PR #270, opened 2022-03-11
feat(kube): private registries support EE-2752

prabhat83:


edge/yaml/yaml.go
```diff
@@ -0,0 +1,143 @@
...
+	credentials := base64.StdEncoding.EncodeToString([]byte(fmt.Sprintf("%s:%s", cred.Username, cred.Secret)))
+	secret := v1Types.Secret{
+		ObjectMeta: v1AMacTypes.ObjectMeta{
+			Name:      secretName,
+			Namespace: namespace,
+		},
+		Data: map[string][]byte{
+			".dockerconfigjson": []byte(fmt.Sprintf(`{
+				"auths": {
+					"https://index.docker.io/v1/": {
```
prabhat83, line 54: This URL should be private registries specific

edge/yaml/yaml.go
```diff
@@ -0,0 +1,143 @@
...
+	credentials := []agent.Credentials{}
+	for _, r := range y.registries {
+		if strings.Contains(imageURL, r.ServerURL) {
+			credentials = append(credentials, r)
+		}
+	}
+	return credentials
+}
+
+func (y *yaml) getImagePullSecret(namespace string, secretName string, cred agent.Credentials) v1Types.Secret {
```
prabhat83, line 44: name could be `generateImagePullSecrets`?

edge/yaml/yaml.go
```diff
@@ -0,0 +1,143 @@
...
+}
+
+// Utility methods
+var re = regexp.MustCompile("[^a-z0-9]+")
+
+func slug(s string) string {
+	return strings.Trim(re.ReplaceAllString(strings.ToLower(s), "-"), "-")
+}
+
+func encodeYAML(yml runtime.Object) (string, error) {
```
prabhat83, line 138: `encodeYML` could be part of `yaml` struct?

edge/yaml/yaml.go
```diff
@@ -0,0 +1,143 @@
...
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/serializer/json"
+
+	"github.com/pkg/errors"
+	"github.com/portainer/agent"
+	v1 "k8s.io/api/apps/v1"
+	"k8s.io/client-go/kubernetes/scheme"
+)
+
+type yaml struct {
```
prabhat83, line 22: this file is basically adding `imagePullSecrets` so the name `yaml` could be confusing?

edge/yaml/yaml.go
prabhat83: maybe using `cred.ServerURL`

PR #271, opened 2022-03-11
fix(chisel): upgrade chisel to v1.7.7 to fix a data race EE-2718

andres-portainer:


chisel/client.go
```diff
@@ -48,10 +48,7 @@ func (client *Client) CreateTunnel(tunnelConfig agent.TunnelConfig) error {
 
 	client.chiselClient = chiselClient
 
-	ctx, cancel := context.WithTimeout(context.Background(), tunnelClientTimeout)
-	defer cancel()
-
-	err = chiselClient.Start(ctx)
+	err = chiselClient.Start(context.Background())
```
deviantony, line 51: I wonder if there is any impact to that change as the context has no timeout anymore?

deviantony, line 51: I wonder if there is any impact to that change as the context has no timeout anymore?

deviantony (commented):
Mostly LGTM

chisel/client.go
andres-portainer: That timeout was not working properly, and I had to remove it because with the new version of Chisel the context cancellation was halting all the communications.

chiptus (approved):
LGTM

PR #284, opened 2022-04-04
feat(edge): auto update process

deviantony:
Related to https://github.com/portainer/agent/pull/284 [EE-3023]

[EE-3023]: https://portainer.atlassian.net/browse/EE-3023?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


cmd/agent/main.go
```diff
@@ -216,9 +223,83 @@ func main() {
 
 	// !Edge
 
+	if options.HealthCheck {
+		log.Println("[INFO] [healthcheck] [message: Agent running on in healthcheck mode. Will exit after running pre-flight checks.]")
```
chiptus, line 227: we may consider moving it to a separate binary

cmd/agent/main.go
```diff
@@ -216,9 +223,83 @@ func main() {
...
+
+		if !options.EdgeMode {
+
+			// TODO: REVIEW
+			// Healthcheck not considered for regular agent in the scope of the agent auto-upgrade POC
```
chiptus, line 232: don't see why not

cmd/agent/main.go
```diff
@@ -216,9 +223,83 @@ func main() {
...
+			goos.Exit(0)
+		}
+
+		// Edge pre-flight checklist
+		// Here we check that the values in the Edge Key can be used by the agent
+		// * The agent can reach out the Portainer instance
+		// * HTTP polling is reachable
+		// * Tunneling is reachable
+
+		edgeKey := edgeManager.GetKey()
```
chiptus, line 244: edgeKey can be nil (if it wasn't set). I guess this would mean `unhealthy`?

cmd/agent/main.go
```diff
@@ -216,9 +223,83 @@ func main() {
...
+
+		// We check that the Portainer instance URL is valid
```
chiptus, line 246: what about checking if the agent api is available? or is that not needed as in that case the process will be closed

chiptus: one case where this might happen is if the API is deadlocked

build/linux/Dockerfile
```diff
@@ -5,4 +5,10 @@ WORKDIR /app
 COPY dist /app/
 COPY static /app/static
 
+# TODO: REVIEW
+# Health check only available on Linux containers for the scope of the POC
```
chiptus, line 9: the only change needed to make it available in other types is copy-pasting L12, right?

cmd/agent/main.go
deviantony: I don't have any particular opinion on that.

cmd/agent/main.go
deviantony: To be discussed with the product team regarding the scope of the piece of work, this was just stated for the POC scope.

cmd/agent/main.go
deviantony: What could cause the agent API to be deadlocked/not avail?

cmd/agent/main.go
deviantony: That's a valid point, I did not consider the manual input of an Edge key (via the Edge UI) in that POC. This is something that we'll need to handle properly.

build/linux/Dockerfile
deviantony: I think that we'll need to change the CMD but I never defined a healthcheck on a Windows container

cmd/agent/main.go
chiptus: I don't know, but what if :)

chiptus:
closed in favor of work done on [EE-3023]


[EE-3023]: https://portainer.atlassian.net/browse/EE-3023?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #288, opened 2022-04-08
feat(build): copy docker config for windows

prabhat83:


hookenz (changes requested):
Looks good but you will also need this line.  This is so that docker can run the credential helper from the path environment.

```
ENV PATH="/app;%PATH%"
```

hookenz (commented):
LGTM :+1:

PR #291, opened 2022-04-12
feat(edge/async): implement three intervals for async polling EE-2731

andres-portainer:


edge/client/interface.go
```diff
@@ -25,7 +25,13 @@ type PollStatusResponse struct {
 	CheckinInterval float64          `json:"checkin"`
 	Credentials     string           `json:"credentials"`
 	Stacks          []StackStatus    `json:"stacks"`
-	AsyncCommands   []AsyncCommand   `json:"commands"` // async mode only
+
+	// Async mode only
```
chiptus, line 30: I think those should go under a nested struct `AsyncSettings` so that it's obvious. this comment is only visible here, it won't be visible for devs in other files

edge/poll_async.go
```diff
@@ -0,0 +1,254 @@
...
+	if interval <= time.Duration(0) {
+		ticker.Stop()
+		return
+	}
+
+	ticker.Reset(interval)
+}
+
+func (service *PollService) failSafe() {
+	zd := time.Duration(0)
```
chiptus: we're using this constant in various places, why not just put it on line 18?

edge/poll_async.go
```diff
@@ -0,0 +1,254 @@
...
+	}
+
+	return service.portainerClient.SetEdgeStackStatus(stackData.ID, responseStatus, errorMessage)
+}
+
+func (service *PollService) processScheduleCommand(command client.AsyncCommand) error {
+	var jobData client.EdgeJobData
+	err := mapstructure.Decode(command.Value, &jobData)
+	if err != nil {
+		log.Printf("[DEBUG] [http,client,portainer] failed to convert %v to edgeStackData", command.Value)
```
chiptus:
```suggestion
		log.Printf("[DEBUG] [http,client,portainer] failed to convert %v to edgeJobData", command.Value)
```

chiptus (approved):
small changes, LGTM otherwise

edge/client/interface.go
andres-portainer: I agree but I'll probably refactor this and other things separately from this PR.

PR #294, opened 2022-04-13
feat(edge): verify certs for revokation [EE-2614]

chiptus:
fix [EE-2614]

[EE-2614]: https://portainer.atlassian.net/browse/EE-2614?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/edge/edge.go
+++ b/edge/edge.go
@@ -13,6 +13,7 @@ import (
 
 	"github.com/portainer/agent"
 	"github.com/portainer/agent/edge/client"
+	"github.com/portainer/agent/edge/revoke"
 	"github.com/portainer/agent/edge/scheduler"
 	"github.com/portainer/agent/edge/stack"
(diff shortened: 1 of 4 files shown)


andres-portainer (approved):

andres-portainer (approved):

deviantony (commented):
Interesting that we seem to copy a package in there - is there no library that can be used to achieve this?

chiptus:
> Interesting that we seem to copy a package in there - is there no library that can be used to achieve this?

@deviantony not that we found. this copied package was ok, but it used too many globals, so I preferred to port it. we probably should have a shared library (libportainer, or just a monorepo)

PR #299, opened 2022-04-21
fix(build): create makefile to correctly build target files EE-2993

hookenz:
Closes [EE-2993](https://portainer.atlassian.net/browse/EE-2993)


build/download_kubectl_binary.sh
```diff
@@ -5,22 +5,6 @@ download_kubectl_binary(){
...
-    ARCH="arm"
-  fi
-
-  if [ "$ARCH" = "aarch64" ]; then
-    ARCH="arm64"
-  fi
-
-  if [ "$ARCH" = "x86_64" ]; then
-    ARCH="amd64"
-  fi
```
hookenz, line 22: PLATFORM and ARCH now come from `go env` output and they are already set to either `arm`, `arm64` or `windows`.

setup.sh
```diff
@@ -40,7 +40,3 @@ else
     download_docker_compose_binary "$PLATFORM" "$ARCH" "$DOCKER_COMPOSE_VERSION"
 fi
 
-source ./dev-scripts/utils.sh
-source ./dev-scripts/compile.sh
-compile_command
-
```
hookenz, line 46: This was a hack to get the CI to build all the binaries (we have two now).  The makefile now builds it

Makefile
```diff
@@ -0,0 +1,23 @@
+# See: https://gist.github.com/asukakenji/f15ba7e588ac42795f421b48b8aede63
+# For a list of valid GOOS and GOARCH values
+# Note: these can be overriden on the command line e.g. `make PLATFORM=<platform> ARCH=<arch>`
+PLATFORM="$(shell go env GOOS)"
```
deviantony, line 4: nice to have these set as vars that you can override 👍

build/download_docker_binary.sh
```diff
@@ -14,6 +14,15 @@ download_docker_binary() {
...
     fi
+
+    if [ "$ARCH" = "arm" ]; then
+        ARCH="armhf"
+    fi
+
+    if [ "$ARCH" = "arm64" ]; then
+        ARCH="aarch64"
+    fi
+
```
hookenz, line 25: These overrides needed ti get the right binary

Makefile
```diff
@@ -0,0 +1,23 @@
...
+agent:
+	@echo "Building agent..."
+	@GOOS=$(PLATFORM) GOARCH=$(ARCH) CGO_ENABLED=0 go build --installsuffix cgo --ldflags '-s'  -o dist/$@ cmd/agent/main.go
+
+docker-credential-portainer:
+	@echo "Building docker-credential-portainer..."
+	@cd cmd/docker-credential-portainer; \
+	GOOS=$(PLATFORM) GOARCH=$(ARCH) CGO_ENABLED=0 go build --installsuffix cgo --ldflags '-s' -o ../../dist/$@
+
+download-binaries:
```
deviantony, line 19: when do you use this one? only independently ? I thought it was required by the agent install?

Makefile
```diff
@@ -0,0 +1,23 @@
+# See: https://gist.github.com/asukakenji/f15ba7e588ac42795f421b48b8aede63
+# For a list of valid GOOS and GOARCH values
+# Note: these can be overriden on the command line e.g. `make PLATFORM=<platform> ARCH=<arch>`
+PLATFORM="$(shell go env GOOS)"
+ARCH="$(shell go env GOARCH)"
+
+
+all: agent docker-credential-portainer download-binaries
+
+agent:
```
deviantony, line 10:
```suggestion
agent: download-binaries
```

See my comment below to automatically call download-binaries when building the agent. You can ignore this if this is actually not required.

build/download_kubectl_binary.sh
```diff
@@ -5,22 +5,6 @@ download_kubectl_binary(){
   local KUBECTL_VERSION=$3
   local KUBECTL_BIN_NAME="kubectl"
 
-  if [ "$PLATFORM" = "win" ]; then
```
deviantony, line 8: Simple question but I'm assuming we're now using the go OS and ARCH references - therefore there is no need for a translation anymore, right?

setup.sh
deviantony: Just make sure it's not breaking other dev local use cases, cc @chiptus as I know he is using this script.

build/download_kubectl_binary.sh
hookenz: Correct.   In some places we need them... but it's simpler now.

Makefile
hookenz: Sometimes you just want to build the agent itself.   e.g. `make agent` and you already have the binaries present.  You don't want to download them again.

I could spend more time changing the download scripts to download only when required but... it's not a big deal.

setup.sh
hookenz: It was my hack.  I added it two nights ago so should be ok

PR #301, opened 2022-04-21
fix(build): create makefile to correctly build target files EE-2993

hookenz:
Closes [EE-2993](https://portainer.atlassian.net/browse/EE-2993)

# Why was this needed?  
The build process for CI was different than what `dev.sh` and `setup.sh` did

CI was calling `setup.sh` and does it's own thing to build the agent

We now have more than one binary being built from the agent source tree and CI didn't cater for it.  So I hacked into setup.sh to build the file.   It was a bit messy and wasn't building things correctly.

This new `Makefile` build process fixes that.

I've also cleaned up the ARCH and PLATFORM options being passed into the setup scripts to make it simpler.  Overrides still exist where required to download the correct binary.

Tested and verified in collaboration with platform team.  We're pushing this out early as CI has changed and all branches will need to rebase to this version on.


stevensbkang (approved):
LGTM

PR #302, opened 2022-04-22
fix(makefile) add phony targets EE-2993

hookenz:
Added targets as .PHONY because they aren't files.  It means if a file called "all", "agent", "clean" etc were added we wouldn't end up with rules that wouldn't execute.


deviantony (approved):
LGTM

PR #316, opened 2022-05-30
feat(edge): add command for edge stack log collection EE-3079

andres-portainer:


edge/client/portainer_edge_async_client.go
```diff
@@ -147,6 +167,36 @@ func (client *PortainerAsyncClient) GetEnvironmentStatus(flags ...string) (*Poll
 				}
 			}
 
+			for _, stack := range client.stackLogCollectionQueue {
+				cs, err := docker.GetContainersWithLabel(fmt.Sprintf("com.docker.compose.project=edge_%s", stack.EdgeStackName))
```
chiptus: this will miss standalone

	ComposeStackNameLabel = "com.docker.compose.project"
	SwarmStackNameLabel   = "com.docker.stack.namespace"

edge/poll_async.go
```diff
@@ -190,7 +192,7 @@ func (service *PollService) processStackCommand(ctx context.Context, command cli
 	var stackData client.EdgeStackData
 	err := mapstructure.Decode(command.Value, &stackData)
 	if err != nil {
-		log.Printf("[DEBUG] [http,client,portainer] failed to convert %v to edgeStackData", command.Value)
+		log.Printf("[DEBUG] [http,client,portainer] failed to convert %v to EdgeStackData", command.Value)
```
chiptus: why don't we log the error too?

chiptus (changes requested):
Looks good. see my comments

edge/client/portainer_edge_async_client.go
andres-portainer: Hmm, what do you mean? I tested this using edge stacks on Docker Standalone and it worked.

chiptus: I meant this will miss swarm. I just saw that you've written we only support standalone, but I think adding the second label will get also swarm

edge/poll_async.go
andres-portainer: Yes, makes sense.

edge/client/portainer_edge_async_client.go
andres-portainer: Okay, I see what you are saying now, sounds good.

edge/client/portainer_edge_client.go
```diff
@@ -251,3 +251,7 @@ func (client *PortainerEdgeClient) ProcessAsyncCommands() error {
 func (client *PortainerEdgeClient) SetLastCommandTimestamp(timestamp time.Time) {
 	return // edge mode only
 }
+
+func (client *PortainerEdgeClient) EnqueueLogCollectionForStack(logCmd LogCommandData) error {
+	return nil
```
chiptus, line 256: is there a reason why we don't enable that for a regular edge?

edge/poll_async.go
```diff
@@ -225,7 +227,7 @@ func (service *PollService) processScheduleCommand(command client.AsyncCommand)
 	var jobData client.EdgeJobData
 	err := mapstructure.Decode(command.Value, &jobData)
 	if err != nil {
-		log.Printf("[DEBUG] [http,client,portainer] failed to convert %v to edgeJobData", command.Value)
+		log.Printf("[DEBUG] [http,client,portainer] failed to convert %v to EdgeJobData", command.Value)
```
chiptus: log error

edge/poll_async.go
```diff
@@ -256,3 +258,17 @@ func (service *PollService) processScheduleCommand(command client.AsyncCommand)
 
 	return nil
 }
+
+func (service *PollService) processLogCommand(command client.AsyncCommand) error {
+	var logCmd client.LogCommandData
+
+	err := mapstructure.Decode(command.Value, &logCmd)
+	if err != nil {
+		log.Printf("[DEBUG] [http,client,portainer] failed to convert %v to LogCommandData", command.Value)
```
chiptus: log error

chiptus (approved):
LG. I did leave some comments

edge/client/portainer_edge_client.go
andres-portainer: We could add it for the regular edge too, but it would have to be a mostly completely new implementation.

PR #324, opened 2022-06-08
fix(compose): rename compose binary for windows EE-3365

hookenz:
Closes [EE-3365](https://portainer.atlassian.net/browse/EE-3365)


dakota-portainer (approved):
LGTM

PR #331, opened 2022-06-16
fix(edge): add image parsing for kube image as docker for private registries EE-3439

prabhat83:


edge/yaml/yaml.go
```diff
@@ -34,7 +35,11 @@ func NewYAML(fileContent string, credentials []agent.RegistryCredentials) *yaml
 func (y *yaml) getRegistryCredentialsByImageURL(imageURL string) []agent.RegistryCredentials {
 	credentials := []agent.RegistryCredentials{}
 	for _, r := range y.registryCredentials {
-		if strings.Contains(imageURL, r.ServerURL) {
+		domain, _, err := getRegistryAndPath(imageURL)
```
dakota-portainer: Any reason why this returns a `path` if we just throw it out and don't use this function anywhere else? Are you thinking you may want the path in the future?

dakota-portainer (approved):
Quick question. Otherwise LGTM.

PR #332, opened 2022-06-16
fix(edge): add image parsing for kube image as docker for private registries EE-3439

prabhat83:


edge/yaml/yaml.go
```diff
@@ -34,7 +35,11 @@ func NewYAML(fileContent string, credentials []agent.RegistryCredentials) *yaml
 func (y *yaml) getRegistryCredentialsByImageURL(imageURL string) []agent.RegistryCredentials {
 	credentials := []agent.RegistryCredentials{}
 	for _, r := range y.registryCredentials {
-		if strings.Contains(imageURL, r.ServerURL) {
+		domain, _, err := getRegistryAndPath(imageURL)
```
dakota-portainer: Any reason why this returns a `path` if we just throw it out and don't use this function anywhere else? Are you thinking you may want the path in the future?

dakota-portainer (approved):
Quick question. Otherwise LGTM.

edge/yaml/yaml.go
```diff
@@ -34,7 +35,11 @@ func NewYAML(fileContent string, credentials []agent.RegistryCredentials) *yaml
...
+		if err != nil {
+			return nil
+		}
+		if strings.Contains(domain, r.ServerURL) || strings.Contains(r.ServerURL, domain) {
```
hookenz: Might be good to explain this logic with a comment.   It basically sanitises the registry imageUrl.

e.g. 
imageUrl = prabhat/nginx   when passed through getRegistryAndPath will return "docker.io" for the domain.
While  imageUrl=quay.io/prabhat/nginx will return "quay.io" for the domain

Second check, we really just need to ensure the domain is in the serverUrl.

hookenz (approved):
Just a minor thing, but otherwise LGTM and approved ahead of time.

edge/yaml/yaml.go
prabhat83: you are right. I copied this function from Matt's work in docker and used it without thinking I guess. Good spotting.

PR #339, opened 2022-06-21
fix(log): log insecure agent warning [EE-3415]

hookenz:
Log a warning if the edge agent is connecting to Portainer without HTTPS

See:  [EE-3415](https://portainer.atlassian.net/browse/EE-3415)

### DIFF
--- a/edge/key.go
+++ b/edge/key.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"log"
+	"net/url"
 	"strconv"
 	"strings"
 
@@ -27,6 +28,11 @@ func (manager *Manager) SetKey(key string) error {
 	edgeKey, err := parseEdgeKey(key)
 	if err != nil {
 		return err
+	}
+
+	u, _ := url.Parse(edgeKey.PortainerInstanceURL)
+	if u.Scheme != "https" {
+		log.Println("[WARN] [edge] [message: This agent has been configured using an insecure connection, which can limit functionality. We recommend updating the agent to use a secure connection.")
 	}
 
 	manager.mu.Lock()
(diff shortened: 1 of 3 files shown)


deviantony (approved):
LGTM

http/server.go
```diff
@@ -91,11 +91,10 @@ func (server *APIServer) Start(edgeMode bool) error {
 		WriteTimeout: 30 * time.Minute,
 	}
 
-	log.Printf("[INFO] [http] [server_addr: %s] [server_port: %s] [secured: %t] [api_version: %s] [message: Starting Agent API server]", server.addr, server.port, config.Secured, agent.Version)
+	log.Printf("[INFO] [http] [server_addr: %s] [server_port: %s] [use_tls: %t] [api_version: %s] [message: Starting Agent API server]", server.addr, server.port, config.UseTLS, agent.Version)
 
-	if edgeMode {
+	if !config.UseTLS {
```
deviantony: I would have kept that condition here as the code in the condition is actually related to Edge mode and not if the server is exposed over HTTP or HTTPS.

hookenz: ok, in that case.

deviantony (approved):
LGTM

prabhat83 (approved):

edge/key.go
```diff
@@ -29,6 +30,11 @@ func (manager *Manager) SetKey(key string) error {
 		return err
 	}
 
+	u, _ := url.Parse(edgeKey.PortainerInstanceURL)
```
prabhat83, line 33: We are not checking for an error here. I think we should do.

hookenz: I don't think in this case it's necessary.   An invalid URL will not work and still produce an error.  But the Portainer UI doesn't let you create an endpoint with a bad URL  So it would have to be hand crafted.  In any case, it still won't run properly and produce errors which is expected.

I'm really just using the function to grab the scheme and if it's not https then log a warning then carry on with the original behavior.

PR #340, opened 2022-06-21
 fix(log): log insecure agent warning [EE-3415]

hookenz:
Log a warning if the edge agent is connecting to Portainer without HTTPS

See:  [EE-3415](https://portainer.atlassian.net/browse/EE-3415)

### DIFF
--- a/edge/key.go
+++ b/edge/key.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"log"
+	"net/url"
 	"strconv"
 	"strings"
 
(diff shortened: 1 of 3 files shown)


deviantony (approved):
LGTM

deviantony (approved):
LGTM

edge/key.go
```diff
@@ -29,6 +30,11 @@ func (manager *Manager) SetKey(key string) error {
 		return err
 	}
 
+	u, _ := url.Parse(edgeKey.PortainerInstanceURL)
```
prabhat83, line 33: Should we check error here?

hookenz: We could, but the front end doesn't allow you to add an endpoint/device with an invalid url.  It'll error out still but just further down.  I'm really just using that function to extract the scheme but I don't care too much if it fails as there is so much error checking in other places where it would still fail if the edge key had a bad url

Issue #341, opened 2022-06-21
Avoid auto switching agent to cluster mode


smartnexus:
I have a Swarm cluster for some network reasons, but I deploy containers in a specific node each time (depending on some requirements). So I want to launch my agent running on Docker standalone but it detects node is part of Swarm cluster, and switches to cluster mode, failing start.

```
2022/06/21 08:01:56 [INFO] [main] [message: Agent running on Docker platform]
2022/06/21 08:01:56 [DEBUG] [main] [Member tags: &{AgentPort:9001 EdgeKeySet:false NodeName:mars DockerConfiguration:{EngineStatus:2 Leader:false NodeRole:2} KubernetesConfiguration:{}}]
2022/06/21 08:01:56 [INFO] [main] [message: Agent running on a Swarm cluster node. Running in cluster mode]
2022/06/21 08:01:56 [DEBUG] [docker] [network_name: bridge] [scope: local] [ingress: false] [message: Skipping invalid container network]
2022/06/21 08:01:56 [WARN] [main] [message: Unable to retrieve agent container IP address, using '0.0.0.0' instead] [error: unable to retrieve the address on which the agent can advertise. Check your network settings]
2022/06/21 08:01:59 [ERROR] [main] [host: tasks.] [message: Unable to retrieve a list of IP associated to the host] [error: lookup tasks. on 172.16.0.2:53: no such host]
```

This is the cmd I am using:
`docker run -it --rm -p 9001:9001 --name portainer_agent -v /var/run/docker.sock:/var/run/docker.sock -v /var/lib/docker/volumes:/var/lib/docker/volumes -e LOG_LEVEL=DEBUG portainer/agent:2.13.1`

I think it will be nice to have an environment variable to disable auto swtiching.

smartnexus:
The key would be to add another condition to the if statement checking options var
https://github.com/portainer/agent/blob/52536ce32c7a2377d35cd19ba5e0301bb759fdd6/cmd/agent/main.go#L80

jamescarppe:
For Portainer to work across a Swarm cluster, even when deploying containers to specific swarm nodes, the agent needs to be running on all the nodes in the cluster. Without it, functionality would be severely limited.

smartnexus:
With this the agent is going to be running in all swarm nodes, but not as a swarm cluster, just a bunch of nodes that are accesible by L3. When the agent runs in swarm mode, deploy is done by "docker stack deploy" and sometimes makes sense to use "docker compose" in just one node

[...]

PR #342, opened 2022-06-27
feat(agent): including new option to disable auto cluster mode

smartnexus:
Mentioned on issue https://github.com/portainer/agent/issues/341


madpeteguy:
I couldn't wait and built version 2.15.1 with the proposed fix: `docker pull ghcr.io/madpeteguy/portainer_agent:2.15.1`
The agent starts but after quick tests I found that it is not able to manage networks or even display a list of existing networks.
Log says:
`http error: Agent management is not available when running the agent on a standalone engine (err=Agent management is disabled) (code=503)`
It looks like disabling swarm mode requires more changes.

PR #355, opened 2022-08-18
feat: update agent dependencies [EE-3895]

DimaSalakhov:


oscarzhou-portainer (approved):
Just bring the attention that the package `golang.org/x/sys` in agent repository is `v0.0.0-20220817070843-5a390386f1f2`, but in portainer-ee is `v0.0.0-20220520151302-bc2c85ada10a`. However, the results are same, which is no vulnerabilities anymore.
So the PR is LGTM

DimaSalakhov:
Yup. In theory all should just work, but I cannot be 100% sure. Have run a few scenarios with this version of the agent to see if anything would fail, so far works fine.

PR #363, opened 2022-09-13
feat(async): add support for more async commands EE-4137

matias-portainer:
closes [EE-4137](https://portainer.atlassian.net/browse/EE-4137)

### DIFF
--- a/docker/docker.go
+++ b/docker/docker.go
@@ -26,10 +26,11 @@ func NewInfoService() *InfoService {
 // GetRuntimeConfigurationFromDockerEngine retrieves information from a Docker environment
 // and returns a map of labels.
 func (service *InfoService) GetRuntimeConfigurationFromDockerEngine() (*agent.RuntimeConfiguration, error) {
-	cli, err := getCli()
+	cli, err := client.NewClientWithOpts(client.FromEnv, client.WithVersion(agent.SupportedDockerAPIVersion))
 	if err != nil {
 		return nil, err
 	}
+	defer cli.Close()
 
 	dockerInfo, err := cli.Info(context.Background())
 	if err != nil {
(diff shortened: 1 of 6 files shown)


docker/docker.go
```diff
@@ -26,11 +26,10 @@ func NewInfoService() *InfoService {
 // GetRuntimeConfigurationFromDockerEngine retrieves information from a Docker environment
 // and returns a map of labels.
 func (service *InfoService) GetRuntimeConfigurationFromDockerEngine() (*agent.RuntimeConfiguration, error) {
-	cli, err := client.NewClientWithOpts(client.FromEnv, client.WithVersion(agent.SupportedDockerAPIVersion))
+	cli, err := getCli()
 	if err != nil {
 		return nil, err
 	}
-	defer cli.Close()
```
andres-portainer: I don't think that `defer` can be removed, the same applies for the other cases.

docker/docker.go
```diff
@@ -136,3 +133,13 @@ func getSwarmConfiguration(config *agent.RuntimeConfiguration, dockerInfo types.
 
 	return nil
 }
+
+func getCli() (*client.Client, error) {
+	cli, err := client.NewClientWithOpts(client.FromEnv, client.WithVersion(agent.SupportedDockerAPIVersion))
+	if err != nil {
+		return nil, err
+	}
+	defer cli.Close()
```
andres-portainer, line 145: This `defer` does not belong here, as it will close the client right after returning it.

andres-portainer (changes requested):
Please add the missing `defer`s.

docker/docker.go
```diff
@@ -136,3 +133,13 @@ func getSwarmConfiguration(config *agent.RuntimeConfiguration, dockerInfo types.
 
 	return nil
 }
+
+func getCli() (*client.Client, error) {
```
andres-portainer: Maybe this function could take a callback `func(cli *client.Client)` so that the other functions don't need to handle the errors nor the resources.

Something like this:
```go
func ContainerStart(name string, opts types.ContainerStartOptions) error {
	return withCli(func(cli *client.Client) error {
		return cli.ContainerStart(context.Background(), name, opts)
	})
}
```

docker/volumes.go
```diff
@@ -0,0 +1,12 @@
...
+import "context"
+
+func VolumeDelete(name string, force bool) error {
+	cli, err := getCli()
+	if err != nil {
+		return err
+	}
+
+	return cli.VolumeRemove(context.Background(), name, force)
+}
```
andres-portainer: Missing newline.

edge/poll_async.go
```diff
@@ -270,3 +277,74 @@ func (service *PollService) processLogCommand(command client.AsyncCommand) error
...
+		err = docker.ContainerDelete(containerCmd.ContainerName, containerCmd.ContainerRemoveOptions)
+	case "kill":
+		err = docker.ContainerKill(containerCmd.ContainerName)
+	}
+
+	if err != nil {
+		log.Printf("[ERROR] [edge] [message: error with '%s' operation on container command] [error: %s]", command.Operation, err)
+	}
+
+	return nil
```
andres-portainer: Shouldn't the error be returned here instead? (the same goes for the cases below)

andres-portainer (approved):

PR #365, opened 2022-09-15
feat(edge/update): remotely update standalone envs [EE-4045]

chiptus:
fixed [EE-4045]

[EE-4045]: https://portainer.atlassian.net/browse/EE-4045?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/edge/client/interface.go
+++ b/edge/client/interface.go
@@ -53,6 +53,8 @@ type VersionUpdate struct {
 	ScheduledTime int64
 	// If need to update
 	Active bool
+	// Update schedule ID
+	ScheduleID int
 }
 
 type setEndpointIDFn func(portainer.EndpointID)
--- a/edge/client/interface.go
+++ b/edge/client/interface.go
@@ -5,13 +5,15 @@ import (
 	"time"
 
 	portainer "github.com/portainer/portainer/api"
+	"github.com/portainer/portainer/api/edgetypes"
 
 	"github.com/portainer/agent"
 )
 
 type EnvironmentStatusOptions struct {
-	DoSnapshot bool
-	DoCommand  bool
+	DoSnapshot          bool
+	DoCommand           bool
+	VersionUpdateStatus *edgetypes.VersionUpdateStatus
 }
 
 type PortainerClient interface {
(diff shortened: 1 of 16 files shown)


edge/poll.go
```diff
@@ -231,6 +250,53 @@ func (service *PollService) poll() error {
...
+
+	service.versionUpdateStatus.ScheduleID = versionUpdate.ScheduleID
+	service.versionUpdateStatus.Error = ""
+	service.versionUpdateStatus.Status = 0
+
+	go func() {
+		service.Stop()
+	}()
+
+	defer service.Start()
```
andres-portainer, line 287: This statement races against `go service.Stop()`, why is it needed to stop it and start it again? The poll won't iterate again until this function returns anyway so I believe the `Start()` and `Stop()` calls can be removed.

edge/poll_async.go
```diff
@@ -58,7 +59,9 @@ func (service *PollService) startStatusPollLoopAsync() {
 
 	log.Debug().Msg("starting Portainer async polling client")
 
-	var snapshotFlag, commandFlag, coalescingFlag bool
+	snapshotFlag := true
+	commandFlag := true
```
andres-portainer, line 63: Why are you setting these flags to `true`? Please keep the original declaration.

edge/poll_async.go
```diff
@@ -81,6 +84,24 @@ func (service *PollService) startStatusPollLoopAsync() {
 		}
 	}
 
+	ping := func() {
```
andres-portainer, line 87: This can't be run as a goroutine, please keep it the way it was.

edge/client/interface.go
```diff
@@ -5,13 +5,20 @@ import (
...
 
+type EnvironmentStatusOptions struct {
+	DoSnapshot          bool
+	DoCommand           bool
+	VersionUpdateStatus *edgetypes.VersionUpdateStatus
+}
+
 type PortainerClient interface {
 	GetEnvironmentID() (portainer.EndpointID, error)
-	GetEnvironmentStatus(flags ...string) (*PollStatusResponse, error)
```
andres-portainer, line 14: The point of using `flags` this way was to avoid having to couple the arguments to a fixed struct that has things that aren't useful in common (for example `DoCommand` is not useful for the regular edge poll).

andres-portainer (changes requested):
Please keep the `startStatusPollLoopAsync()` as it was.

PR #379, opened 2022-10-03
feat(edge/stacks): support scheduling [EE-3023]

chiptus:
[EE-3023]

[EE-3023]: https://portainer.atlassian.net/browse/EE-3023?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/agent.go
+++ b/agent.go
@@ -215,7 +215,6 @@ var (
 )
 
 const (
-
 	// APIVersion represents the version of the agent's API.
 	APIVersion = "2"
 	// DefaultAgentAddr is the default address used by the Agent API server.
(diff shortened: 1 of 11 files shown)


matias-portainer (approved):

andres-portainer:
Now that the schedules could fire a long time after they were saved I think `SetLastCommandTimestamp()` probably needs to be changed so that it always increases the timestamp instead of just replacing it blindly, to avoid traveling back in time.

andres-portainer (approved):

PR #395, opened 2022-11-07
fix(edge/docker): allow to deploy stack with private registry on async edge [EE-4531]

oscarzhou-portainer:
closes [EE-4531]

[EE-4531]: https://portainer.atlassian.net/browse/EE-4531?

### DIFF
--- a/edge/client/portainer_edge_async_client.go
+++ b/edge/client/portainer_edge_async_client.go
@@ -438,7 +438,34 @@ func (client *PortainerAsyncClient) DeleteEdgeStackStatus(edgeStackID int) error
 
 // GetEdgeStackConfig retrieves the configuration associated to an Edge stack
 func (client *PortainerAsyncClient) GetEdgeStackConfig(edgeStackID int) (*agent.EdgeStackConfig, error) {
-	return nil, nil // unused in async mode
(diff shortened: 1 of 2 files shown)


edge/client/portainer_edge_async_client.go
```diff
@@ -438,7 +438,34 @@ func (client *PortainerAsyncClient) DeleteEdgeStackStatus(edgeStackID int) error
 
 // GetEdgeStackConfig retrieves the configuration associated to an Edge stack
 func (client *PortainerAsyncClient) GetEdgeStackConfig(edgeStackID int) (*agent.EdgeStackConfig, error) {
-	return nil, nil // unused in async mode
+	requestURL := fmt.Sprintf("%s/api/endpoints/%d/edge/stacks/%d", client.serverAddress, client.getEndpointIDFn(), edgeStackID)
```
andres-portainer: Async mode **MUST NOT** make any extra requests to Portainer, all the information exchange needs to happen via the async polling loop, which uses `/endpoints/edge/async`. This is a strict requirement.

andres-portainer (changes requested):
This PR needs to be reworked so it sends the required information as part of the edge stack command.

andres-portainer (approved):

PR #400, opened 2022-11-21
feat(edge/update): inject tls certificate info in nomad agent [EE-3220]

oscarzhou-portainer:
closes [EE-3220]

This PR does not need to test before another PR https://github.com/portainer/agent/pull/379 is merged

[EE-3220]: https://portainer.atlassian.net/browse/EE-3220?

### DIFF
--- a/edge/stack/stack.go
+++ b/edge/stack/stack.go
@@ -1,13 +1,16 @@
 package stack
 
 import (
+	"bytes"
 	"context"
 	"fmt"
 	"os"
 	"path/filepath"
+	"strings"
 	"sync"
 	"time"
 
+	"github.com/cbroglie/mustache"
 	"github.com/portainer/agent"
 	"github.com/portainer/agent/edge/client"
 	"github.com/portainer/agent/edge/yaml"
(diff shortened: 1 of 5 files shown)


edge/stack/stack.go
```diff
@@ -158,6 +158,7 @@ func (manager *StackManager) processStack(stackID int, version int) error {
 	}
 	if manager.engineType == EngineTypeNomad {
 		fileName = fmt.Sprintf("%s.hcl", stack.Name)
+
```
chiptus: remove empty line

nomad/nomad_deployer.go
```diff
@@ -169,3 +174,35 @@ func compareJobs(old, new *nomadapi.Job) bool {
 
 	return false
 }
+
+// addTLSInfo will inject Nomad TLS certificate info to updater job if the TLS certificate is provided
+func addTLSInfo(job *nomadapi.Job) {
+	// Always make sure that job.ID matches with the job ID in nomad mustache template on server side
+	// Suggest to use a job name with format "portainer-updater-{uuid}""
+	targetJobName := "portainer-updater"
+	if *job.ID == targetJobName && *job.TaskGroups[0].Name == targetJobName && job.TaskGroups[0].Tasks[0].Name == targetJobName {
```
chiptus: do we need to use `*job` or `job`? because they have both in this line

chiptus: also, might be good to check that there are TaskGroups and Tasks (i.e `len()>0`)

nomad/nomad_deployer.go
```diff
@@ -169,3 +174,35 @@ func compareJobs(old, new *nomadapi.Job) bool {
 
 	return false
 }
+
+// addTLSInfo will inject Nomad TLS certificate info to updater job if the TLS certificate is provided
+func addTLSInfo(job *nomadapi.Job) {
+	// Always make sure that job.ID matches with the job ID in nomad mustache template on server side
+	// Suggest to use a job name with format "portainer-updater-{uuid}""
```
chiptus: this is specifically for portainer-updater, maybe we should have a boolean on the edge job? `addTLSInfo`?

chiptus (changes requested):
Looks good. a few things that need to be checked

nomad/nomad_deployer.go
oscarzhou-portainer: > do we need to use `*job` or `job`? because they have both in this line

Yes. It is because the type of `job.ID` and `job.TaskGroups[0].Name` is `*string` and the type of `job.TaskGroups[0].Tasks[0].Name` is `string`

> also, might be good to check that there are TaskGroups and Tasks (i.e `len()>0`)

Good point

nomad/nomad_deployer.go
oscarzhou-portainer: I think it is a good idea. Even though there is the job name `portainer-updater` check, it is still not 100% accurate.  More importantly, it is less secure.

chiptus:
the following commits should be dropped: 
Screenshot_20221204_142506

as they were applied in the standalone remote update

nomad/nomad_deployer.go
chiptus: do you want to apply this or add it in the future?

nomad/nomad_deployer.go
```diff
@@ -169,3 +174,46 @@ func compareJobs(old, new *nomadapi.Job) bool {
...
+	targetJobName := "portainer-updater"
+	if *job.ID == targetJobName &&
+		len(job.TaskGroups) > 0 &&
+		*job.TaskGroups[0].Name == targetJobName &&
+		len(job.TaskGroups[0].Tasks) > 0 &&
+		job.TaskGroups[0].Tasks[0].Name == targetJobName {
+		task := job.TaskGroups[0].Tasks[0]
```
chiptus:
````
if !shouldApplyTLS(job) {
  return
}

...

```

nomad/nomad_deployer.go
oscarzhou-portainer: As we added the boolean flag for identifying the nomad edge job, I think it is unnecessary to implement the proposed idea. I will remove the comments.

nomad/nomad_deployer.go
oscarzhou-portainer: fixed

chiptus (approved):
LGTM

nomad/nomad_deployer.go
```diff
@@ -169,3 +185,62 @@ func compareJobs(old, new *nomadapi.Job) bool {
 
 	return false
 }
+
+// addNomadDefaultEnv injects environment varibles inherited from Nomad environment
+func addNomadDefaultEnv(job *nomadapi.Job) {
+	task := job.TaskGroups[0].Tasks[0]
```
andres-portainer, line 191: Maybe it's a good idea to check the length of `TaskGroups` and `TaskGroups[0].Tasks` before this line?

oscarzhou-portainer: It's a good consideration. The check was added in this function but later extracted out which is in line 241-245

PR #405, opened 2022-12-06
fix(makefile): fix gopls error, makefile help

hookenz:
Some improvements to the Makefile and also removing a `gopls` error due to the go.mod in `docker-credential-portainer` which has one at the top level.

Added `go.work` file

Closes: [DTD-121](https://portainer.atlassian.net/browse/DTD-121)


Makefile
```diff
@@ -12,22 +12,24 @@ agent=agent
...
 
-download-binaries:
+download-binaries: ## Download dependant binaries
 	@./setup.sh $(PLATFORM) $(ARCH)
 
-clean:
+clean: ## Remove all build and download artifacts
 	@rm -f dist/*
 
+help:
```
chiptus: add this as a default target?

cmd/docker-credential-portainer/go.mod
```diff
@@ -1,5 +0,0 @@
-module github.com/portainer/docker-credential-portainer
```
chiptus, line 1: as discussed, I think it's better to have this as a separate module and use `go.work`

Makefile
hookenz: I'd like to, need to check CI process

cmd/docker-credential-portainer/go.mod
hookenz: But it's still a cmd so follows the recommended practice of having commands inside cmd directory.
By moving it to the top level, we're breaking that practise and it'll make navigating sources harder if we add more binaries in other directories at the top level.   The way I've fixed it seems right IMO.

Also I think go.work is for modules you're importing to your project not the other way around.  It allows go sub-commands to operate on multiple modules at once.  So you do not need to add a hack to go.mod with replace anymore.

hookenz: Ok, it seems to work.  I think we should use `go.work` but keep the command in the sources beneath the `cmd` directory.

dakota-portainer (approved):
Simple and clean. I like comments you added in the makefile.

PR #421, opened 2023-02-02
feat(update/registry): support to remote update from private registry [EE-3460]

oscarzhou-portainer:
closes [EE-3460]

[EE-3460]: https://portainer.atlassian.net/browse/EE-3460?


edge/stack/stack.go
```diff
@@ -151,14 +152,20 @@ func (manager *StackManager) processStack(stackID int, version int) error {
...
+		}
+	} else if manager.engineType == EngineTypeKubernetes {
 		fileName = fmt.Sprintf("%s.yml", stack.Name)
 		if len(stackConfig.RegistryCredentials) > 0 {
 			yml := yaml.NewYAML(fileContent, stackConfig.RegistryCredentials)
 			fileContent, _ = yml.AddImagePullSecrets()
 		}
-	}
-	if manager.engineType == EngineTypeNomad {
+	} else if manager.engineType == EngineTypeNomad {
```
chiptus: why need `else`? engine type can be both

edge/yaml/yaml.go
```diff
@@ -167,3 +193,60 @@ func encodeYAML(yml runtime.Object) (string, error) {
...
+	if !ok {
+		return "", errors.Wrap(err, fmt.Sprintf("Cannot find the service: %s", serviceName))
+	}
+
+	log.Info().Int("number", len(envs)).Msg("environment variable")
+	for k, v := range envs {
+		service.Environment = append(service.Environment, fmt.Sprintf("%s=%s", k, v))
+	}
+
+	// Needs to reorder the elements in the env array, as Golang built-in array do not guarantee
```
chiptus: why do we care about the order?

edge/yaml/yaml.go
```diff
@@ -167,3 +193,60 @@ func encodeYAML(yml runtime.Object) (string, error) {
...
+
+func addEnvsForSpecificService(fileContent, serviceName string, envs map[string]string) (string, error) {
+	var compose Compose
+	err := libYaml.Unmarshal([]byte(fileContent), &compose)
+	if err != nil {
+		return "", errors.Wrap(err, "Error while unmarshalling the docker compose file content")
+	}
+
+	if !validateComposeFile(&compose, serviceName) {
+		return "", errors.New("Fail to validate the compose file content")
```
chiptus:
```suggestion
		return "", errors.New("Failed to validate the compose file content")
```

edge/yaml/yaml.go
```diff
@@ -152,6 +165,19 @@ func (y *yaml) AddImagePullSecrets() (string, error) {
 	return strings.Join(ymlFiles, "---\n"), nil
 }
 
+func (y *yaml) AddCredentialsAsEnvForSpecificService(serviceName string) (string, error) {
```
chiptus: I think it's a bit confusing that we have docker and kubernetes in the same file and the same type.  I suggest moving the new logic to `yaml/docker.go` and creating a new type for this `dockerYaml`

edge/stack/stack.go
oscarzhou-portainer: You meant that engine type can be `EngineTypeKubernetes` and `EngineTypeNomad` at the same time?

chiptus: no, I meant that this change is not needed. else is not needed here

```
if a=0 {
 ... do something
} else if a=1 {
	... do something else {
} else if a=2 {...}
```

if like
```
if a=0 {
 ... do something
} 
if a=1 {
	... do something else {
} 
if a=2 {...}
```

a switch case is probably the best use

oscarzhou-portainer: Three `if` will check the condition three times no matter whether either of them matches or not. `if ... else if ..` will skip the rest `if` once the condition is true. In this case, `switch case break` has the same effect.

edge/yaml/yaml.go
oscarzhou-portainer: I found that allocating the memory in advance will guarantee the order of the element in the slice. For example,
```
	if service.Environment == nil {
		service.Environment = make([]string, 0)
	}
	for k, v := range envs {
		service.Environment = append(service.Environment, fmt.Sprintf("%s=%s", k, v))
	}
```
So I removed `sort()` here

edge/stack/stack.go
chiptus: it's just a readability thing, not making things faster (if this difference is our speed problem then we're in a good place). I'd say a switch case is the right use case here

chiptus (approved):
LGTM

PR #437, opened 2023-03-06
fix(updater/nomad): remove the ghost nomad updater container [EE-5105]

oscarzhou-portainer:
closes [EE-5105]

[EE-5105]: https://portainer.atlassian.net/browse/EE-5105?

### DIFF
--- a/internals/updates/docker.go
+++ b/internals/updates/docker.go
@@ -6,15 +6,19 @@ import (
 	"fmt"
 
 	"github.com/docker/docker/api/types"
+	"github.com/docker/docker/api/types/filters"
 	"github.com/docker/docker/client"
 	"github.com/portainer/agent"
 )
 
 type DockerUpdaterCleaner struct {
+	updateID int
 }
 
-func NewDockerUpdaterCleaner() *DockerUpdaterCleaner {
-	return &DockerUpdaterCleaner{}
+func NewDockerUpdaterCleaner(updateID int) *DockerUpdaterCleaner {
+	return &DockerUpdaterCleaner{
+		updateID: updateID,
+	}
 }
 
 func (du *DockerUpdaterCleaner) Clean(ctx context.Context) error {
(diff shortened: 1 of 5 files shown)


internals/updates/update.go
```diff
@@ -0,0 +1,40 @@
...
+	"time"
+
+	"github.com/rs/zerolog/log"
+)
+
+type GhostUpdaterCleaner interface {
+	Clean(ctx context.Context) error
+}
+
+func Remove(ctx context.Context, updateID int, fn GhostUpdaterCleaner) error {
```
chiptus: `fn` sounds like function. why not use a func?
```suggestion
func Remove(ctx context.Context, cleanFn func(ctx) ) error {
```

then you can call it just 
```updates.Remove(ctx, updaterCleaner.Clean)```

Comment (chiptus) [reply internals/updates/update.go]: for the updateId, may I suggest pass it when initalizing the cleaner? as this is the condition on which you create it?

File: internals/updates/docker.go [hunk 2]
```diff
@@ -0,0 +1,59 @@
...
+				if container.NetworkSettings != nil {
+					for _, networkSetting := range container.NetworkSettings.Networks {
+						err = cli.NetworkRemove(ctx, networkSetting.NetworkID)
+						if err != nil {
+							return fmt.Errorf("failed to remove network: %s", err.Error())
+						}
+					}
+				}
+			} else if container.State == "running" {
+				foundRunningContainer = true
```
chiptus: why not just return the error here?

internals/updates/docker.go
```diff
@@ -0,0 +1,59 @@
...
+	}
+	defer cli.Close()
+
+	foundRunningContainer := false
+	containers, err := cli.ContainerList(ctx, types.ContainerListOptions{All: true})
+	if err != nil {
+		return fmt.Errorf("failed to list containers: %s", err.Error())
+	}
+	for _, container := range containers {
+		_, ok := container.Labels["io.portainer.hideStack"]
```
chiptus: didn't we decide to use a different label? something like `io.portainer.updater=true`?  `hideStack` can be used to hide other stacks, so it might return wrong containers. 

also, why not call `cli.ContainerList` with the label filter?

internals/updates/update.go
```diff
@@ -0,0 +1,40 @@
...
+	// Make three attempts to allow the updater container to exit on its own.
+	// This is important because if the updater container is forcibly removed
+	// before it exits naturally, it may skip removing the previous agent's
+	// container by the updater container, resulting in a container name conflict
+	// during the next remote update.
```
chiptus, line 20: what about just like in docker, check if the updater job is done, if it's not done, continue and wait more?

oscarzhou-portainer: When using Docker, we need to rename the new container manually once the updater job removes the previous one. This is because there is always a container with a unique name, `portainer_agent`. However, with Nomad as an orchestrator, the old container is automatically removed after the new container is created. There is no unique container name, but instead a random string is appended to the name `portainer_agent`, resulting in `portainer_agent_xxxxxx`.

internals/updates/update.go
oscarzhou-portainer: > `fn` sounds like function. why not use a func?
> 
> then you can call it just `updates.Remove(ctx, updaterCleaner.Clean)`

I forgot to change the parameter name "fn" after modifying the function definition from anonymous function to interface, even though my intention was to pass the object.

internals/updates/docker.go
oscarzhou-portainer: Although we had previously made the decision to update the server from CE to BE, the agent update was merged before that decision was made. As a result, I will update here accordingly to ensure consistency in the implementation.

internals/updates/docker.go
oscarzhou-portainer: In the event that there are multiple containers that match, it is necessary to evaluate all of them. While it is preferable to have only one container with the label `io.portainer.updater=true`, this solution is not problematic.

xAt0mZ (approved):

internals/updates/nomad.go
```diff
@@ -0,0 +1,62 @@
...
+
+	// Remove the job
+	jobID := "portainer-updater"
+
+	job, _, err := client.Jobs().Info(jobID, nil)
+	if err != nil {
+		log.Debug().
+			Err(err).
+			Msg("failed to find nomad job portainer-updater")
+		return errors.Wrap(err, "failed to find nomad job portainer-updater")
```
chiptus: if we didn't find the job, it's ok, no? I think in this case just `return nil`

internals/updates/update.go
```diff
@@ -0,0 +1,48 @@
...
+import (
+	"context"
+	"time"
+
+	"github.com/rs/zerolog/log"
+)
+
+type GhostUpdaterCleaner interface {
+	Clean(ctx context.Context) error
+	UpdateID() int
```
chiptus, line 12: it seems like the actual updateId isn't used by the cleaner. it's just there for log in the `Remove` func. why not just pass it to `Remove`?

chiptus: passing it to the cleaner makes it look like they will clean the updater for this specific id, when it cleans all ghost updaters.

oscarzhou-portainer: That was my original implementation: https://github.com/portainer/agent/pull/437/commits/9a1f71771a61ef836eb95715534c87c759430d9f#

internals/updates/nomad.go
oscarzhou-portainer: Right. I think I didn't change back after debugging.

internals/updates/update.go
chiptus: sorry, you are right. I made this comment https://github.com/portainer/agent/pull/437#discussion_r1126349216

just didn't see that it's actually not required. it's actually not really required for the Remove fn. we can and maybe should look for those ghosts also when agent is not an updated agent

oscarzhou-portainer: It could be a good idea. I am simply curious about the possibility of a new agent (not an updated agent) starting while the ghost updater container exists.

PR #447, opened 2023-03-15
feat(update/registry): remote update from ECR private registry with cert [EE-5164] 

oscarzhou-portainer:
closes [EE-5164]

[EE-5164]: https://portainer.atlassian.net/browse/EE-5164?


edge/yaml/docker_compose.go
```diff
@@ -57,30 +51,77 @@ func addEnvsForSpecificService(fileContent, serviceName string, envs map[string]
...
+		if err != nil {
+			return "", err
+		}
+
+		envs["REGISTRY_USED"] = "1"
+		// hardcode username for aws ecr registry
+		// @https://docs.aws.amazon.com/cli/latest/reference/ecr/get-login-password.html#examples
+		envs["REGISTRY_USERNAME"] = "AWS"
+		envs["REGISTRY_PASSWORD"] = c.Secret
+		envs["REGISTRY_PASSWORD"] = c.Secret
```
chiptus: remove duplicate line

edge/yaml/docker_compose_test.go
```diff
@@ -1,109 +1,84 @@
...
-			name: "valid docker compose file content",
-			input: `version: "3"
-services:
-invalid_service:`,
-			expected: ``,
-			wantErr:  true,
+			name:      "empty image name",
+			imageName: "",
+			expected:  "",
+			err:       errors.New("No image name provided"),
```
chiptus, line 99: is something like `http://localhost:3000/image:tag` a possible registry? and what about `localhost:3000/image:tag`?

edge/yaml/docker_compose.go
```diff
@@ -57,30 +51,77 @@ func addEnvsForSpecificService(fileContent, serviceName string, envs map[string]
...
+
+				envs["REGISTRY_USED"] = "1"
+				envs["REGISTRY_USERNAME"] = cred.Username
+				envs["REGISTRY_PASSWORD"] = cred.Secret
+				break
+			}
+		}
+	}
+
+	return updateServiceWithEnv(compose, serviceName, envs)
```
chiptus, line 97: maybe we should think builder pattern. so we call this:
```
ym := yaml.NewDockerComposeYaml(....); // here we already have parsed ym as `ym.parsed` or `ym.data`
ym.WithCredentails("updater")
ym.WithEnv("updater", envs)
newFile := ym.String()
```

wdyt?

edge/aws/aws.go
```diff
@@ -71,3 +76,18 @@ func authenticateAgainstIAMRA(awsConfig *agent.AWSConfig) (*iamra.CredentialProc
 
 	return &credentialProcessOutput, nil
 }
+
+func ExtractAwsConfig(options *agent.Options) *agent.AWSConfig {
+	if os.IsValidAWSConfig(options) {
```
chiptus: maybe move IsValidAWSConfig here?

edge/stack/stack.go
```diff
@@ -512,8 +514,8 @@ func (manager *StackManager) buildDeployerParams(stackData client.EdgeStackData,
 
 	switch manager.engineType {
 	case EngineTypeDockerStandalone, EngineTypeDockerSwarm:
-		if len(stackData.RegistryCredentials) > 0 && strings.HasPrefix(stackData.Name, "edge-update-schedule") {
-			yml := yaml.NewDockerComposeYAML(fileContent, stackData.RegistryCredentials)
+		if (len(stackData.RegistryCredentials) > 0 || manager.awsConfig != nil) && strings.HasPrefix(stackData.Name, "edge-update-schedule") {
```
chiptus: tech debt: change the last condition from `strings.HasPrefix(stackData.Name, "edge-update-schedule")` to `stackData.AddCredentials`

edge/yaml/docker_compose_test.go
chiptus: just to be clear, I'm not sure if those are valid registries, but if they are, we need to take care of them

chiptus (commented):
mostly suggestions, need to remove duplicate line and see that we take care of all registry types.

edge/yaml/docker_compose_test.go
oscarzhou-portainer: I'm surprised to see how many valid image URL formats there are. To ensure we have adequate test coverage, I will be adding more unit test cases.

edge/yaml/docker_compose.go
oscarzhou-portainer: If we're only working with Docker Compose YAML files, I don't think we need to create a pattern right away. Alternatively, we could consider developing a pattern that covers both Docker Compose YAML and K8s YAML in a separate project to make the workflow more clear and organized.

edge/stack/stack.go
oscarzhou-portainer: Can you please explain this more?
This registry credential injection here is only for remote update from private registry

edge/yaml/docker_compose_test.go
chiptus: maybe we can use https://github.com/distribution/distribution/blob/v2.7.1/reference/reference.go#L189 ?

edge/stack/stack.go
chiptus: it's very specific. what if we change the name of the stack? this should be an edge stack property and not some magic string check

chiptus (approved):

edge/registry/server.go
```diff
@@ -37,7 +38,7 @@ func NewEdgeRegistryHandler(edgeManager *edge.Manager, awsConfig *agent.AWSConfi
 func (handler *Handler) LookupHandler(rw http.ResponseWriter, r *http.Request) *httperror.HandlerError {
 	stackManager := handler.EdgeManager.GetStackManager()
 	if stackManager == nil {
-		return &httperror.HandlerError{http.StatusInternalServerError, "Unable to retrieve stack manager", errors.New("Stack manager is not available")}
+		return httperror.InternalServerError("Unable to retrieve stack manager", errors.New("Stack manager is not available"))
```
chiptus, line 41: please apply these changes to develop version

chiptus (changes requested):
apply changes to develop

edge/registry/server.go
oscarzhou-portainer: Someone already made the changes to the develop branch before I could submit my pull request.

chiptus (approved):
LGTM

PR #478, opened 2023-06-18
feat(edge/stacks): check for stacks status [EE-5554]

chiptus:
related to [EE-5554]

[EE-5554]: https://portainer.atlassian.net/browse/EE-5554?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


edge/stack/stack.go
```diff
@@ -521,6 +610,10 @@ func buildEnvVarsForDeployer(envVars []portainer.Pair) []string {
 }
 
 func (manager *StackManager) deleteStack(ctx context.Context, stack *edgeStack, stackName, stackFileLocation string) {
+	manager.mu.Lock()
```
andres-portainer, line 616: Why is this lock needed here?

exec/kubernetes_deploy_status.go
```diff
@@ -0,0 +1,18 @@
...
+	"context"
+
+	libstack "github.com/portainer/portainer/pkg/libstack"
+)
+
+func (service *KubernetesDeployer) WaitForStatus(ctx context.Context, name string, status libstack.Status) <-chan string {
+	resultCh := make(chan string)
+
+	go func() {
+		resultCh <- ""
```
andres-portainer: There is no need to write to the channel here, the `close()` will achieve the same effect.

nomad/nomad_deployer.go
```diff
@@ -255,3 +256,14 @@ func isUpdateJob(job *nomadapi.Job) bool {
 		len(job.TaskGroups[0].Tasks) > 0 &&
 		job.TaskGroups[0].Tasks[0].Name == targetJobName
 }
+
+func (service *Deployer) WaitForStatus(ctx context.Context, name string, status libstack.Status) <-chan string {
+	resultCh := make(chan string)
+
+	go func() {
+		resultCh <- ""
```
andres-portainer: Same as https://github.com/portainer/agent/pull/478/files#r1254914543.

exec/docker_swarm_stack_status.go
```diff
@@ -0,0 +1,170 @@
...
+	"github.com/portainer/agent/docker"
+	libstack "github.com/portainer/portainer/pkg/libstack"
+	"github.com/rs/zerolog/log"
+)
+
+func (service *DockerSwarmStackService) WaitForStatus(ctx context.Context, name string, status libstack.Status) <-chan string {
+	errorMessageCh := make(chan string)
+
+	go func() {
+		timeout := time.After(5 * time.Minute)
```
andres-portainer: Why do we need the timeout if we (can) have it in the context?

exec/docker_swarm_stack_status_test.go
```diff
@@ -0,0 +1,148 @@
...
+
+	statusCh := deployer.WaitForStatus(ctx, stackName, requiredStatus)
+	select {
+	case result := <-statusCh:
+		if result == "" {
+			return requiredStatus, "", nil
+		}
+
+		return libstack.StatusError, result, nil
+	case <-ctx.Done():
```
andres-portainer: This channel read contends with the the same one in `deployer.WaitForStatus()`, only one can succeed, I suggest removing it from here.

exec/docker_swarm_stack_status.go
```diff
@@ -0,0 +1,170 @@
...
+		for {
+			select {
+			case <-timeout:
+				return
+			case <-ctx.Done():
+				return
```
andres-portainer: Shouldn't we write to `resultCh` before returning?

andres-portainer (approved):

edge/stack/stack.go
chiptus: because we're changing the stack status. I thought it's needed

exec/docker_swarm_stack_status.go
chiptus: right, it's left from before I added timeout to context

exec/docker_swarm_stack_status_test.go
chiptus: :+1:

edge/stack/stack.go
andres-portainer: It was needed for this line that is not there anymore
```go
delete(manager.stacks, edgeStackID(stack.ID))
```

I'm not seeing any mutation to the `manager` state here, but I could be missing it.

chiptus: `stack *edgeStack` is an object from the manager state, so changing it means changing the manager state. isn't it?

PR #479, opened 2023-06-19
feat(edge-stack): relative path support for edge stack EE-5521

cmenginnz:


edge/stack/stack.go
```diff
@@ -115,6 +117,53 @@ func (manager *StackManager) UpdateStacksStatus(pollResponseStacks map[int]int)
...
+
+	return nil
+}
+
+func getStackFileFolder(stack *edgeStack) string {
+	stackIDStr := strconv.Itoa(int(stack.ID))
+
+	folder := path.Join(agent.EdgeStackFilesPath, stackIDStr)
+	if IsRelativePathStack(stack) {
+		folder = path.Join(stack.FilesystemPath, agent.ComposePathPrefix, stackIDStr)
```
oscarzhou-portainer: use `filepath` package for both ln159 and ln161

edge/stack/stack.go
```diff
@@ -289,6 +315,14 @@ func (manager *StackManager) performActionOnStack(queueSleepInterval time.Durati
 			return
 		}
 
+		if IsRelativePathStack(stack) {
+			dst := path.Join(stack.FilesystemPath, agent.ComposePathPrefix)
```
oscarzhou-portainer: `filepath`

edge/stack/stack.go
```diff
@@ -515,106 +559,84 @@ func (manager *StackManager) DeleteStack(ctx context.Context, stackData edge.Sta
...
+	if err != nil {
+		return err
+	}
+
+	if !deleteStack {
+		err = filesystem.PersistDir(stack.FileFolder, stackPayload.DirEntries)
+		if err != nil {
+			return err
+		}
+	}
```
oscarzhou-portainer, line 650: Should line 624-639 be placed above line 566?

oscarzhou-portainer (changes requested):
In terms of the proposed solution, I have some concerns that I will include in the ticket.

Overall, the PR itself LGTM. There are also other few places using `path` which are not highlighted. Would recommend to use `filepath`.

edge/stack/stack.go
cmenginnz: fixed

edge/stack/stack.go
cmenginnz: fixed

edge/stack/stack.go
cmenginnz: fixed. The modification is applied to a cloned stack now.

exec/docker_copy.go
```diff
@@ -0,0 +1,74 @@
...
+				fmt.Sprintf("%s:%s", composeDestination, composeDestination),
+			},
+		},
+		nil,
+		nil,
+		containerName,
+	)
+}
+
+func copyToContainer(assetPath, src, containerID, dst string) error {
```
chiptus: what about using `dockerCli.CopyToContainer`? https://github.com/moby/moby/blob/master/client/container_copy.go#L33

docker/containers.go
```diff
@@ -5,9 +5,31 @@ import (
...
+
+	err := withCli(func(cliInner *client.Client) error {
+		cli = cliInner
+		return nil
+	})
+	if err != nil {
+		return container.CreateResponse{}, err
+	}
+
+	return cli.ContainerCreate(context.Background(), config, hostConfig, networkingConfig, platform, containerName)
```
chiptus: this feels a bit unsafe and beats the purpose of `withCli` fn, isn't cli closed inside `withCli`? is there any reason why not to run these lines of code inside withCli?

exec/docker_copy.go
chiptus: also IMO should be inside `docker` package

docker/containers.go
cmenginnz: fixed

exec/docker_copy.go
cmenginnz: moved to docker package. 
`dockerCli.CopyToContainer` requires tar as input. Using the docker command is simpler here.

oscarzhou-portainer (approved):
LGTM

PR #497, opened 2023-07-25
feat: use alternative ECR login logic

deviantony:
This PR provides an alternative approach to AWS ECR authentication when using the IAM RA configuration.

### DIFF
--- a/edge/aws/aws.go
+++ b/edge/aws/aws.go
@@ -2,18 +2,20 @@ package aws
 
 import (
 	"context"
+	"encoding/base64"
 	"errors"
+	"strings"
 
 	"github.com/aws/aws-sdk-go-v2/config"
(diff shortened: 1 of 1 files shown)


edge/aws/aws.go
```diff
@@ -30,34 +32,45 @@ func DoAWSIAMRolesAnywhereAuthAndGetECRCredentials(serverURL string, awsConfig *
...
-		return nil, ErrNoCredentials
+	token, err := base64.StdEncoding.DecodeString(*data.AuthorizationToken)
+	if err != nil {
+		log.Err(err).Msg("unable to decode ECR authorization token")
+		return nil, err
 	}
 
+	tokenParts := strings.Split(string(token), ":")
+	username := tokenParts[0]
+	password := tokenParts[1]
```
andres-portainer, line 68: What happens if the password has a `:`?

PR #566, opened 2024-01-11
chore(tilt): Tilt agent scripts [EE-7239]

hookenz:
[EE-7239]

Tilt integration to make development with the Portainer Agent much easier in different environment types.

[EE-7239]: https://portainer.atlassian.net/browse/EE-7239?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.env.example
```diff
@@ -0,0 +1,6 @@
+TZ=Pacific/Auckland
```
chiptus, line 1: shouldn't be set IMO

Tiltfile
```diff
@@ -0,0 +1,46 @@
+local_resource(
+  'build_agent',
+  cmd="make agent",
+  deps=['*.go', 
+    'agent.go',
```
chiptus, line 5: there isn't a way to depend on all go files? something like `**/*.go`?

docker-compose.yaml
```diff
@@ -0,0 +1,20 @@
...
+
+version: "3.9"
+services:
+  agent:
+    image: portainer-agent
+    container_name: portainer_edge_agent_dev
+    volumes:
+      - /var/run/docker.sock:/var/run/docker.sock
+      - /var/lib/docker/volumes:/var/lib/docker/volumes
+      - /usr/share/zoneinfo:/usr/share/zoneinfo
```
chiptus, line 12: what's this? timezone?

docker-compose.yaml
```diff
@@ -0,0 +1,20 @@
...
+      - /:/host
+      - portainer_agent_data:/data
+    restart: always
+    env_file:
```
chiptus, line 16: I think this is not necessary, used by default

tilt/README.md
```diff
@@ -0,0 +1,37 @@
+# Using Tilt (tilt.dev) with the Portainer Agent
+
+## Directory Contents
+This directory contains some tiltfiles for use with Tilt.
+
+## Getting started for developers
```
chiptus, line 6: does it allow debug? putting breakpoints?

docker-compose.yaml
hookenz: Because I was running portainer outside of a container and the agent ran inside a container it meant that they were running with different timezone information.   Therefore when you pushed an update script it wouldn't run.  Yes, normally you wouldn't need the TZ set

.env.example
hookenz: If the API is running locally outside of a container it will pick up the timezone.
If you need the agent also to run in the same timezone you would need to set this and pass in the timezone db.

Tiltfile
hookenz: No.

tilt/README.md
hookenz: It's completely possible.  But I didn't get around to doing that.  It requires attaching delv. 

Details here:
https://www.kriskelly.me/go-dlv-tilt-debugging/

PR #577, opened 2024-02-01
chore(edgestacks): remove unnecessary error handling EE-6663

andres-portainer:


chiptus (approved):
the actual change (removal of error from Stop), was a bit hidden between all the renames.
LG

PR #661, opened 2024-06-13
fix(edge): parse the response body with failed request [EE-7244]

oscarzhou-portainer:
closes [EE-7244]

[EE-7244]: https://portainer.atlassian.net/browse/EE-7244?


edge/client/portainer_edge_client.go
```diff
@@ -51,10 +51,6 @@ type NonOkResponseError struct {
 	msg string
 }
 
-func newNonOkResponseError(msg string) *NonOkResponseError {
-	return &NonOkResponseError{msg: msg}
-}
-
```
andres-portainer, line 57: You cannot remove this, we rely on that error type in other parts of the code.

edge/client/portainer_edge_async_client.go
```diff
@@ -415,14 +415,7 @@ func (client *PortainerAsyncClient) executeAsyncRequest(payload AsyncRequest, po
...
 	if resp.StatusCode != http.StatusOK {
-		errorData := parseError(resp)
-		logError(resp, errorData)
-
-		if errorData != nil {
-			return nil, errors.New(errorData.Message + ": " + errorData.Details)
-		}
-
-		return nil, errors.New("short poll request failed")
+		return nil, logPollingError(resp, "AsyncEdgeAgentExecuteAsyncRequest", fmt.Sprintf("AsyncEdgeAgent [%d] failed to execute async request", client.getEndpointIDFn()))
```
andres-portainer: Please try to avoid using `Sprintf()` for this purpose, it's much better to use the structured logging functions that zlog provides, in this case something like `.Int("endpoint_id", endpointID)`.

edge/client/error_response.go
```diff
@@ -2,40 +2,28 @@ package client
...
-	Message string
-}
-
-func parseError(resp *http.Response) *errorData {
-	errorData := &errorData{}
-
-	err := json.NewDecoder(resp.Body).Decode(&errorData)
-	if err != nil {
-		log.Debug().CallerSkipFrame(1).
+func logPollingError(resp *http.Response, ctxMsg, errMsg string) error {
```
andres-portainer: I don't think we should mix logging with error handling.

edge/client/error_response.go
```diff
@@ -2,40 +2,28 @@ package client
...
-	if errorData == nil {
-		return
-	}
-
-	log.Debug().CallerSkipFrame(1).
-		Str("error_response_message", errorData.Message).
-		Str("error_response_details", errorData.Details).
-		Int("status_code", resp.StatusCode).
-		Msg("poll request failure")
+	log.
```
andres-portainer: Shouldn't we use `CallerSkipFrame()` here as before?

andres-portainer (changes requested):
Please see the comments.

edge/client/portainer_edge_client.go
oscarzhou-portainer: The type was not removed, only the constructor function was removed, because I found that it is not used.

andres-portainer: That is technically true but the constructor function is not used anymore because you removed it in this PR. :rofl: 

You can't do that, we need to return that error because we depend on it in other places, please see: https://github.com/portainer/agent/blob/676d62a86b33071fe51172c42fca74c80df3ac95/edge/poll.go#L217-L225

edge/client/portainer_edge_async_client.go
```diff
@@ -415,14 +415,18 @@ func (client *PortainerAsyncClient) executeAsyncRequest(payload AsyncRequest, po
...
-
-		if errorData != nil {
-			return nil, errors.New(errorData.Message + ": " + errorData.Details)
+		ctxMsg := "AsyncEdgeAgentExecuteAsyncRequest"
+		errMsg := "AsyncEdgeAgent failed to execute async request"
+		if err := decodeNonOkayResponse(resp, ctxMsg); err != nil {
+			log.
+				Error().Err(err.Err).
+				Str("context", ctxMsg).
+				Str("response message", err.Message).
```
deviantony, line 424:
```suggestion
				Str("response_message", err.Message).
```
We need to have a consistent naming for variables, I suggest that we use snake case for this. I will update the guidelines with that.

deviantony (commented):
I didn't had the time to do a full review yet but I just wanted to drop a quick comment regarding consistency across logging var names.

PR #662, opened 2024-06-14
chore(code): clean up the code EE-7251

andres-portainer:


edge/stack/stack.go
```diff
@@ -581,10 +587,9 @@ func (manager *StackManager) deployStack(ctx context.Context, stack *edgeStack,
 	manager.mu.Lock()
 	defer manager.mu.Unlock()
 
-	stack.DeployCount += 1
+	stack.DeployCount++
```
xAt0mZ, line 590: I find `stack.DeployCount += 1` easier to read than `rawstack.DeployCount++` (likely because I'm used to read ++ only in for loops)

Same goes for all other += 1 changed to ++

edge/registry/server.go
```diff
@@ -70,27 +70,29 @@ func (handler *Handler) LookupHandler(rw http.ResponseWriter, r *http.Request) *
...
-
-			if strings.HasSuffix(u.Hostname(), "docker.io") {
-				key = "docker.io"
-			} else {
-				key = u.Hostname()
-			}
+
+	if len(credentials) == 0 {
+		return response.Empty(rw)
+	}
```
xAt0mZ, line 76: While this is an early return that can improve a tiny bit the execution, the flow below will yield the same result. I'm not 100% sure that we should keep this early return

docker/docker_copy.go
```diff
@@ -75,7 +74,6 @@ func removeAndCopy(src, dst string, stackID int, stackName, assetPath string, ne
 
 func removeUnpackerContainer(unpackerContainer container.CreateResponse) error {
 	err := ContainerDelete(unpackerContainer.ID, container.RemoveOptions{})
-
 	if err != nil {
 		log.Error().
 			Str("ContainerID", unpackerContainer.ID).
```
xAt0mZ: More explicit, and keep err def/eval in line with the codebase.

```go
func removeUnpackerContainer(unpackerContainer container.CreateResponse) error {
  if err := ContainerDelete(unpackerContainer.ID, container.RemoveOptions{}); err != nil {
    log.Error().
      Str("ContainerID", unpackerContainer.ID).
      Msg("Failed to remove unpacker container")
    return err
  }
  return nil
}
```

xAt0mZ (approved):

edge/registry/server.go
andres-portainer: It's just to reduce nesting, while it's not a big deal in this case, we have others that are multiple levels deep, so I'm trying to reduce the "bad examples".

PR #670, opened 2024-07-05
fix(edgestacks): keep trying to set the edge stack status on network or server errors EE-7246

andres-portainer:


deviantony (approved):
LGTM

PR #682, opened 2024-08-09
fix(updater): normalize the agent status and name [EE-7388]

oscarzhou-portainer:
Ref: BE-10984

closes [EE-7388]

[EE-7388]: https://portainer.atlassian.net/browse/EE-7388?

### DIFF
--- a/internals/updates/docker.go
+++ b/internals/updates/docker.go
@@ -4,17 +4,27 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"strconv"
+	"strings"
 
+	"github.com/docker/docker/api/types"
 	"github.com/docker/docker/api/types/container"
 	"github.com/docker/docker/api/types/filters"
+	"github.com/docker/docker/client"
 	"github.com/portainer/agent/docker"
+	"github.com/rs/zerolog/log"
 )
 
 type DockerUpdaterCleaner struct {
 	updateID int
 }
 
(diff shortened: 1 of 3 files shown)


internals/updates/docker.go
```diff
@@ -52,17 +62,120 @@ func (du *DockerUpdaterCleaner) Clean(ctx context.Context) error {
...
 }
+
+func updateAgentInfoIfNeeds(ctx context.Context, updateID int) error {
+	cli, err := docker.NewClient()
+	if err != nil {
+		return err
+	}
+	defer cli.Close()
+
+	possibleImagePrefixes := []string{"portainer/agent", "portainerci/agent"}
```
xAt0mZ: You cannot use a predefined list here. Portainer server has a `EDGE_UPDATE_AGENT_IMAGE_PREFIX ` and `EDGE_UPDATE_AGENT_IMAGE_VERSION` env vars that can be used to deploy any image name

oscarzhou-portainer: > You cannot use a predefined list here. Portainer server has a `EDGE_UPDATE_AGENT_IMAGE_PREFIX ` and `EDGE_UPDATE_AGENT_IMAGE_VERSION` env vars that can be used to deploy any image name

good catch. we can use other info to locate the agent, such as labels.

xAt0mZ (approved):

PR #725, opened 2024-10-04
fix(edge): reset stacks in mem when agents not trusted [BE-11110]

oscarzhou-portainer:
Ref: BE-11110

### DIFF
--- a/edge/client/interface.go
+++ b/edge/client/interface.go
@@ -87,11 +87,12 @@ type StackStatus struct {
 
 type setEndpointIDFn func(portainer.EndpointID)
 type getEndpointIDFn func() portainer.EndpointID
+type resetEdgeStacksFn func()
 
 // NewPortainerClient returns a pointer to a new PortainerClient instance
-func NewPortainerClient(serverAddress string, setEIDFn setEndpointIDFn, getEIDFn getEndpointIDFn, edgeID string, edgeAsyncMode bool, agentPlatform agent.ContainerPlatform, metaFields agent.EdgeMetaFields, httpClient *edgeHTTPClient) PortainerClient {
+func NewPortainerClient(serverAddress string, setEIDFn setEndpointIDFn, getEIDFn getEndpointIDFn, resetESFn resetEdgeStacksFn, edgeID string, edgeAsyncMode bool, agentPlatform agent.ContainerPlatform, metaFields agent.EdgeMetaFields, httpClient *edgeHTTPClient) PortainerClient {
 	if edgeAsyncMode {
-		return NewPortainerAsyncClient(serverAddress, setEIDFn, getEIDFn, edgeID, agentPlatform, metaFields, httpClient)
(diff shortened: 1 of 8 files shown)


edge/client/interface.go
```diff
@@ -87,11 +87,12 @@ type StackStatus struct {
 
 type setEndpointIDFn func(portainer.EndpointID)
 type getEndpointIDFn func() portainer.EndpointID
+type resetEdgeStacksFn func()
```
andres-portainer: I'm not sure this is really needed, it doesn't add any clarity.

edge/poll.go
```diff
@@ -19,6 +19,7 @@ import (
 )
 
 const tunnelActivityCheckInterval = 30 * time.Second
+const waitToBeAssociated = 0
```
andres-portainer: The name is not accurate, the `0` value means that the agent is using a global edge key so the endpoint ID will be automatically created, independently of the waiting room.

edge/client/interface.go
```diff
@@ -87,11 +87,12 @@ type StackStatus struct {
...
 
 // NewPortainerClient returns a pointer to a new PortainerClient instance
-func NewPortainerClient(serverAddress string, setEIDFn setEndpointIDFn, getEIDFn getEndpointIDFn, edgeID string, edgeAsyncMode bool, agentPlatform agent.ContainerPlatform, metaFields agent.EdgeMetaFields, httpClient *edgeHTTPClient) PortainerClient {
+func NewPortainerClient(serverAddress string, setEIDFn setEndpointIDFn, getEIDFn getEndpointIDFn, resetESFn resetEdgeStacksFn, edgeID string, edgeAsyncMode bool, agentPlatform agent.ContainerPlatform, metaFields agent.EdgeMetaFields, httpClient *edgeHTTPClient) PortainerClient {
```
andres-portainer: A slightly simpler approach could be to fix the inconsistency we have here: https://github.com/portainer/agent/blob/develop/edge/client/portainer_edge_async_client.go#L413:L424 so it matches this code: https://github.com/portainer/agent/blob/develop/edge/client/portainer_edge_client.go#L170:L179, then we can reset the endpoint ID (and stacks manager) directly in `poll_async.go` as `poll.go` does (https://github.com/portainer/agent/blob/develop/edge/poll.go#L217:L225). This way we avoid the need for a new callback.

andres-portainer (approved):

edge/client/interface.go
oscarzhou-portainer: I agree. Initially, I was unsure if it would disrupt the flow since I assumed there must have been a reason for injecting the callback function into the client. It's great to have your confirmation, and I'll go ahead and simplify the flow a bit.

edge/poll.go
oscarzhou-portainer: Do you have any suggestions for a better name? '0' doesn't really convey any meaning and could lead to confusion.

oscarzhou-portainer: Have renamed it `globalKeyInUse`

